PR #43, opened 2022-10-24
Relayer node service test

arcticfloyd1984:


packages/account-abstraction/src/HttpRpcClient.ts
```diff
@@ -3,18 +3,21 @@ import { ethers } from 'ethers'
...
+import { HttpMethod, sendRequest } from './utils/httpRequests'
 export class HttpRpcClient {
   private readonly userOpJsonRpcProvider: JsonRpcProvider
+  chainId: number;
 
   constructor (
     readonly bundlerUrl: string,
     readonly entryPointAddress: string,
-    readonly chainId: number
+    chainId: number
```
livingrockrises, line 14: why this change has been done?

PR #50, opened 2022-10-31
feat: social login ui added

AmanRaj1608:


packages/web3-auth/src/UI.tsx
```diff
@@ -0,0 +1,129 @@
...
+      />
+      <div
+        style={{
+          display: 'flex',
+          justifyContent: 'center',
+          alignItems: 'center',
+          position: 'relative',
+          color: '#FFFFFF',
+          flexDirection: 'column'
+        }}
```
asheerrizvi, line 81: Let's create a stylesheet and use classes.

packages/web3-auth/src/UI.tsx
```diff
@@ -0,0 +1,129 @@
...
+          Create a wallet to continue
+        </p>
+      </div>
+      <div>
+        <button onClick={() => props.socialLogin.login()} style={googleCardStyle}>
+          <img
+            src={'https://live-nft-hosted-assets.s3.ap-south-1.amazonaws.com/google_logo.svg'}
+            style={{ marginRight: 14 }}
+          />
+          <span style={buttonTextSpan}>Continue with Google</span>
```
asheerrizvi, line 114: Move these objects to classes as well.

packages/web3-auth/src/UI.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+const footerBigText = {
+  fontStyle: 'normal',
+  fontWeight: '700',
+  fontSize: '16.1351px',
+  lineHeight: '21px',
+  color: '#535353'
+} as React.CSSProperties
+
+const UI = (props: any) => {
```
asheerrizvi, line 65: Can we specify what props this component expects instead of using `any`?

packages/web3-auth/src/UI.tsx
AmanRaj1608: have created the stylesheets in https://github.com/bcnmy/biconomy-client-sdk/tree/webpack-changes
the webpack doesn't builds the css properly, let's merge this and look if we want to switch to stylesheet or not.

livingrockrises (approved):
LGTM

PR #51, opened 2022-10-31
added messaging sdk

kunal047:


packages/relayer/src/rest-relayer.ts
```diff
@@ -71,6 +72,7 @@ export class RestRelayer implements Relayer {
     this.setRelayerNodeEthersProvider(chainId)
 
     const { multiSendCall } = context // multisend has to be multiSendCallOnly here!
+    let finalRawRx
```
livingrockrises, line 75: this is for code optimisation so we eventually make one sendRequest?

PR #52, opened 2022-11-03
custom paymaster + smart account config

livingrockrises:


packages/account-abstraction/src/BiconomyPaymasterAPI.ts
```diff
@@ -1,22 +1,15 @@
...
-  // Might maintain API key at smart account level
-  constructor(
-    readonly apiUrl: string,
-    readonly dappAPIKey: string,
-    readonly payMasterAddress: string
-  ) {
+import { IPaymasterAPI } from '@biconomy-sdk/core-types'
+
+export class BiconomyPaymasterAPI implements IPaymasterAPI {
+  constructor(readonly apiUrl: string, readonly dappAPIKey: string) {
```
tomarsachin2271, line 7: apiUrl should be renamed to signingServiceUrl

packages/core-types/src/smart-account.types.ts
```diff
@@ -9,19 +9,21 @@ import { MultiSendCallOnlyContract } from './contracts/MultiSendCallOnlyContract
...
   signType: SignTypeMethod
-  providerUrlConfig?: ProviderUrlConfig[]
+  // dappAPIKey?: string
+  networkConfig: NetworkConfig[]
   entryPointAddress?: string
-  bundlerUrl?: string
-  paymasterAddress?: string
-  signingServiceUrl: string
+  biconomySigningServiceUrl?: string
+  // bundlerUrl?: string
```
tomarsachin2271, line 24: Uncomment bundlerUrl here

packages/core-types/src/smart-account.types.ts
```diff
@@ -9,19 +9,21 @@ import { MultiSendCallOnlyContract } from './contracts/MultiSendCallOnlyContract
 import { SmartWalletContract } from './contracts/SmartWalletContract'
 import { GasLimit } from './transaction.types'
 import { Signer } from 'ethers'
+import { IPaymasterAPI } from 'account-abstraction-types'
 
 export interface SmartAccountConfig {
   activeNetworkId: ChainId // same
   supportedNetworksIds: ChainId[] // Network[] chainId: CbainId, rpcUrl?: string
```
tomarsachin2271, line 16: Add environment field which can have values "testnet" or "mainnet" and maintain two default values, 1 for testnet and 1 for mainnet. Be default it should be "mainnet"

livingrockrises: can you share more details? i missed this part of the discussion

packages/core-types/src/smart-account.types.ts
livingrockrises: done

packages/account-abstraction/src/BiconomyPaymasterAPI.ts
livingrockrises: done

packages/core-types/src/smart-account.types.ts
tomarsachin2271: The default values for the config which we have are of test networks but we should have one default configuration for main networks also.... and to do that we can have one 'environment' field in SmartAccountConfig which can have values "test" or "mainnet".. 

Like we have for [Hyphen](https://docs.biconomy.io/products/hyphen-instant-cross-chain-transfers/sdk#2.-importing-and-instantiation)

So if developer set this field as "test" then we'll use the defaultConfig value which has test networks values.. and vice versa.

livingrockrises: got it. let me create another PR

PR #64, opened 2022-11-11
added onHashChanged event

kunal047:


packages/account-abstraction/src/ERC4337EthersProvider.ts
```diff
@@ -178,17 +178,6 @@ export class ERC4337EthersProvider extends BaseProvider {
               receipt: tx.receipt
             })
             resolve(receipt)
```
kunal047, line 180: this is removed as the event is already being listened in the signer class

livingrockrises (approved):

PR #95, opened 2022-12-12
Release/1.0.0

livingrockrises:
preparing for release 0.0.5

list of PRs merged into branch release/0.0.5


packages/web3-auth/src/SocialLogin.tsx
```diff
@@ -54,25 +54,31 @@ class SocialLogin {
...
-    network: 'mainnet' | 'testnet' = 'testnet'
-  ) {
+  async init(socialLoginDTO?: Partial<SocialLoginDTO>) {
+    const finalDTO: SocialLoginDTO = {
+      chainId: '0x1',
+      whitelistUrls: {},
+      network: 'testnet'
+    }
+    if (socialLoginDTO) {
+      if (socialLoginDTO.chainId) finalDTO.chainId = socialLoginDTO.chainId
```
tomarsachin2271, line 64: Under what circumstances this chainId needs to be overridden?? and how does 'testnet' value of network field changes the EOA account generated by social login sdk?? 
@AmanRaj1608 @livingrockrises

packages/web3-auth/src/types/Web3AuthConfig.ts
```diff
@@ -1,4 +1,9 @@
 export type DefaultSocialLoginConfig = {
-  defaultChainId: string
   backendUrl: string
 }
+
+export type SocialLoginDTO = {
+  chainId: string
+  whitelistUrls: { [P in string]: string }
```
tomarsachin2271, line 7: What does this type means? { [P in string]: string }

tomarsachin2271:
Also in this release lets move the console logs to a utility method and use a debug flag to log the messages in console.

packages/web3-auth/src/types/Web3AuthConfig.ts
AmanRaj1608: The whitelistUrls expects an array of domain to signature which is passed to web3Auth on initialisation. So basically array of key string to value string.
Example 
```json
{
	"http://localhost:3000": "signature.....",
	"http://newdomain.com": "signature.....",
}
```

packages/web3-auth/src/SocialLogin.tsx
AmanRaj1608: - On web3Auth Initialisation they expect a chainId cause they generate a provider and signer related to that chainId only. 
- after sign in with soicals, the web3auth servers to which it redirects changes on what we pass, which changes the EOA.

tomarsachin2271: Would this change in chainId change the generated EOA address as well?

AmanRaj1608: no chainId is used to give back the provider only

AmanRaj1608: network values can be


image

packages/account-abstraction/src/SmartAccountAPI.ts
```diff
@@ -114,6 +114,10 @@ export class SmartAccountAPI extends BaseWalletAPI {
    * @param info
    */
   async createUnsignedUserOp(info: TransactionDetailsForUserOp): Promise<UserOperation> {
+    // not ideal as it will unnecessary increase callGasLimit, unless appliedGasLimit is sent reasonable
+    /*if ((await this.checkWalletDeployed())) {
```
tomarsachin2271, line 118: Remove these comments if not needed.

packages/common/src/Logger.ts
```diff
@@ -0,0 +1,12 @@
+
+  /**
+   * Single method to be used for logging purpose.
+   *
+   * @param {any} message Message to be logged
+   */
+   export function _logMessage(message: any) {
+    // Method would be called ifLogsEnabled is true
```
tomarsachin2271, line 8: Logger will be a class with static method log with a static state `isLogsEnabled`, which is set by the main package constructor based on users input by calling the static method `enableLogs` on the class.

```
export class Logger {
    static isLogsEnabled: boolean = false;

    static log(message: any) {
      if(this.isLogsEnabled && console.log) {
        console.log(message);
      }
    }

    static enableLogs(_isLogsEnabled: boolean) {
      this.isLogsEnabled = _isLogsEnabled;
    }
}

```
Also i don't see the usage of this method anywhere in the code yet.

packages/smart-account/src/SmartAccount.ts
```diff
@@ -307,24 +315,41 @@ class SmartAccount extends EventEmitter {
...
+    let transactionRequest : TransactionRequest = {...transaction}
 
+    const isDeployed = await this.contractUtils.isDeployed(
+      chainId,
+      this.DEFAULT_VERSION,
+      this.address
+    )
+
+    if(!isDeployed || isDelegate) {
+       customData.appliedGasLimit = ethers.constants.Two.pow(24) // estimateGas for execFromEntryPoint
```
tomarsachin2271, line 342: Why hardcoded value here?

livingrockrises: so userOp callGasLimit is estimation on execFromEntryPoint method on wallet.

```
// Review: doesn't work if wallet is not already deployed.
    const callGasLimit =
      parseNumber(detailsForUserOp.gasLimit) ??
      (await this.provider.estimateGas({
        from: this.entryPoint.address,
        to: this.getWalletAddress(),
        data: callData
      }))
```
      
 so if wallet is not deployed it just estimates and returns 21000 + data cost  then it fails
 
 Further changes is marked for testing / brainstorming and to be launched in next release before gasless goes live on 
 mainnet.

packages/account-abstraction/src/SmartAccountAPI.ts
livingrockrises: done

packages/common/src/Logger.ts
livingrockrises: acknowledged and noted for logger ticket scope as discussed

PR #113, opened 2023-01-25
fix: smartAccount

AmanRaj1608:
# Description

Fix the SmartAccount package to accept ethers signer class from backend.


livingrockrises (approved):
LGTM

PR #121, opened 2023-02-20
Stable changes scw 1.0.15

talhamalik883:
This PR include following major smart contract changes

1. Addition of getNonce(1) function for forward flow
2. Addition of nonce() for AA flow
3. Removal of entryPoint and handler params from deployCounterFactualWallet function
4. Removal of batchId from ExecTransaction


packages/core-types/src/contracts/SmartWalletContract.ts
```diff
@@ -18,18 +18,17 @@ export interface SmartWalletContract {
...
   execTransaction(
     transaction: ExecTransaction,
-    batchId: number,
     feeRefundData: IFeeRefundV1_0_0,
     signatures: string
   ): any
   /* eslint-disable  @typescript-eslint/no-explicit-any */
   execTransaction(
     transaction: ExecTransaction,
-    batchId: number,
```
livingrockrises, line 32: what if we tomorrow deploy a new version that has batchId field?

packages/smart-account/src/SmartAccount.ts
```diff
@@ -411,8 +411,8 @@ class SmartAccount extends EventEmitter {
       sender: this.address,
       target: this.address,
       nonce: gasTankNonce,
-      callData: execTransaction.data || '',
-      callGasLimit: BigNumber.from(500000), // will be updated below
+      callData: execTransactionData || '',
```
livingrockrises, line 414: what is the difference here execTransactionData and execTransaction.data?
how it was working earlier as I don't see other change

packages/smart-account/src/SmartAccount.ts
```diff
@@ -452,18 +452,18 @@ class SmartAccount extends EventEmitter {
...
-        fallbackUserOp.callData,
-        chainId
-      )
-      fallbackUserOp.callGasLimit = BigNumber.from(callGasLimit)
-      this._logMessage('callGasLimit')
-      this._logMessage(callGasLimit)
-    } catch (error) {
-      this._logMessage('Error in estimating callGasLimit')
-    }
+    // try {
```
livingrockrises, line 455: I feel this should not get commented out. I will spend some time to check entire flow post merge

packages/smart-account/src/SmartAccount.ts
livingrockrises: I see

packages/smart-account/src/SmartAccount.ts
```diff
@@ -496,42 +496,27 @@ class SmartAccount extends EventEmitter {
       config: state,
       context: this.getSmartAccountContext(chainId)
     }
-    const gasLimit = transactionDto.transaction.gasLimit
```
livingrockrises, line 499: Is it necessary to remove this? it's only removed for fallback relayer or before sending it to any relayer?

livingrockrises (approved):
do check comments for future reference. approving

PR #129, opened 2023-03-20
Add eip1271 functions

filmakarov:
It's a **draft** of PR for the ease of comparing the changes

The goal is to add the support for the following 3 functions:

- signMessageOnchain
- getMessageHash
- isValidSignature

despite all 3 of them should be called like smartAccount.method, they are not defined in the SmartAccount.sol (or it's parents) , but they are defined in the FallbackHandler and SignMessageLib, so they are called via call/delegatecall.

[https://github.com/bcnmy/scw-contracts/blob/f9f2b057906841fb4b9825cdc9e85a05b4d03a[…]tracts/smart-contract-wallet/handler/DefaultCallbackHandler.sol](https://github.com/bcnmy/scw-contracts/blob/f9f2b057906841fb4b9825cdc9e85a05b4d03a84/contracts/smart-contract-wallet/handler/DefaultCallbackHandler.sol#L38-L57)
https://github.com/bcnmy/scw-contracts/blob/f9f2b057906841fb4b9825cdc9e85a05b4d03a84/contracts/smart-contract-wallet/libs/SignMessageLib.sol#L23-L50


talhamalik883 (approved):

packages/common/package.json
```diff
@@ -39,6 +39,7 @@
   "dependencies": {
     "@account-abstraction/contracts": "^0.5.0",
     "@biconomy/core-types": "^1.0.0",
+    "@biconomy/node-client": "^1.0.0",
```
livingrockrises, line 42: do we need to / must we import node client in common?

packages/common/src/Types/IEntryPoint.ts
```diff
@@ -14,32 +14,22 @@ import type {
...
-  TypedListener,
-  OnEvent,
-  PromiseOrValue,
-} from "./common";
+import { FunctionFragment, Result, EventFragment } from "@ethersproject/abi";
+import { Listener, Provider } from "@ethersproject/providers";
+import { TypedEventFilter, TypedEvent, TypedListener, OnEvent } from "./common";
 
 export type UserOperationStruct = {
-  sender: PromiseOrValue<string>;
```
livingrockrises, line 32: can leave it as it is?

livingrockrises (approved):

PR #130, opened 2023-03-20
feat: improve logs

AmanRaj1608:
- Use one Logger to console all logs.
- Use the env variable to enable the logs
- Use [consoleColors.js](https://gist.github.com/abritinthebay/d80eb99b2726c83feb0d97eab95206c4) to improve logs readability.

### DIFF
--- a/packages/common/src/Logger.ts
+++ b/packages/common/src/Logger.ts
@@ -5,46 +5,44 @@
  */
 class Logger {
   // By default, the logger is not in debug mode.
-  private readonly isDebug: boolean = false
-
-  constructor() {
-    // Check if the debug flag is set in the environment variables
-    if (process.env.BICONOMY_SDK_DEBUG === 'true') {
-      this.isDebug = true
-    }
-  }
+  static isDebug: boolean =
+    process.env.BICONOMY_SDK_DEBUG === 'true'
+      ? true
+      : process.env.REACT_APP_BICONOMY_SDK_DEBUG === 'true'
+      ? true
+      : false
 
   /**
    * \x1b[0m is an escape sequence to reset the color of the text
    * All color codes used - 31 - Red, 33 - Yellow, 34 - Blue, 35 - Magenta, 36 - Cyan
-   * log - Magenta[time] Cyan[message]: Blue[value]
-   * warn - Magenta[time] Yellow[WARN]: Cyan[message] Yellow[value]
-   * error - Magenta[time] Red[ERROR]: Cyan[message] Red[value]
+   * log -   Magenta[time]               Cyan[message]:  [value]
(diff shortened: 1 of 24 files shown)


AmanRaj1608:
logs now
   * log - Magenta[time] Cyan[message]: Blue[value]
   * warn - Magenta[time] Yellow[WARN]: Cyan[message] Yellow[value]
   * error - Magenta[time] Red[ERROR]: Cyan[message] Red[value]

image

packages/relayer/src/RestRelayer.ts
```diff
@@ -177,14 +176,12 @@ export class RestRelayer implements IRelayer {
...
-              hash: txHash
-            })}`
-          )
+          this.logger.log('Tx Hash generated message received at client ', {
+            transactionId: txId,
+            hash: txHash
+          })
 
-          console.log(`Receive time for transaction id ${txId}: ${Date.now()}`)
+          this.logger.log('Receive time for transaction id', Date.now())
```
livingrockrises, line 184: why txId is removed?

livingrockrises: i think this line is not required, can remove it

packages/smart-account/src/SmartAccount.ts
```diff
@@ -172,7 +159,7 @@ class SmartAccount extends EventEmitter {
     } else if (Provider.isProvider(signerOrProvider)) {
       this.signer = new SmartAccountSigner(signerOrProvider)
     } else {
-      this._logMessage('signer or provider is not valid')
+      this.logger.log('signer or provider is not valid')
```
livingrockrises, line 162: should be logger.error

livingrockrises (commented):
as a dev how can change the env?

packages/common/src/Logger.ts
```diff
@@ -1,11 +1,50 @@
...
-    console.log(message)
+class Logger {
+  // By default, the logger is not in debug mode.
+  static isDebug: boolean =
+    process.env.BICONOMY_SDK_DEBUG === 'true'
+      ? true
+      : process.env.REACT_APP_BICONOMY_SDK_DEBUG === 'true'
+      ? true
+      : false
+
```
AmanRaj1608, line 14: Here checking two variables, either developer can se `BICONOMY_SDK_DEBUG` or `REACT_APP_BICONOMY_SDK_DEBUG` true.

This is because in backend we can set the env variable and use the env variable, but for frontend developer have two choises
-  Use dotenv to resolve the local .env variable
- Use the REACT_APP_ prefix variable as in `create-react-app` by default [reads all the env variables which starts with REACT_APP_](https://stackoverflow.com/a/48378498/11097431)

talhamalik883 (approved):

PR #137, opened 2023-03-29
Signing service negative case (allow wallet self sponsor)

livingrockrises:
feat/strictSponsorshipMode and fix/BiconomyPaymasterAPI edge case + dev notes + update type

# Description

* added strictSponsorshipMode field in ClientConfig
* made constructor of BiconomyPaymasterAPI to just take one argument as object
* based on strictSponsorshipMode (default value currently false and dapps can't pass) and if signingService fails we will return '0x' so the wallet can self sponsor
* if account still doesnt have funds it should fail from bundler service with proper error code / message
[ signing service never returns other statusCode than 200 or message and always throws which is something we can look into @arcticfloyd1984  @talhamalik883 ] 

Fixes # (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] This change requires a documentation update

# How Has This Been Tested?

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/account-abstraction/src/BiconomyPaymasterAPI.ts
+++ b/packages/account-abstraction/src/BiconomyPaymasterAPI.ts
@@ -1,24 +1,22 @@
 import { resolveProperties } from '@ethersproject/properties'
 import { UserOperation } from '@biconomy/core-types'
 import { HttpMethod, sendRequest } from './utils/httpRequests'
-import { IPaymasterAPI } from '@biconomy/core-types'
+import { IPaymasterAPI, PaymasterConfig } from '@biconomy/core-types'
 import { Logger } from '@biconomy/common'
 
 /**
  * Verifying Paymaster API supported via Biconomy dahsboard to enable Gasless transactions
  */
+// TODO: possibly rename to BiconomyVerifyingPaymasterAPI
 export class BiconomyPaymasterAPI implements IPaymasterAPI {
-  constructor(readonly signingServiceUrl: string, readonly dappAPIKey: string) {
-    this.signingServiceUrl = signingServiceUrl
-    this.dappAPIKey = dappAPIKey
+  paymasterConfig: PaymasterConfig
+
+  constructor(paymasterConfig: PaymasterConfig) {
+    this.paymasterConfig = paymasterConfig
   }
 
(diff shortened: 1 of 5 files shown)


packages/smart-account/src/SmartAccount.ts
```diff
@@ -260,6 +260,9 @@ class SmartAccount extends EventEmitter {
         new ethers.providers.JsonRpcProvider(providerUrl),
         {
           dappAPIKey: clientConfig.dappAPIKey || '',
+          // Review: default false
+          // could come from global set config or method level when we implement fee mode
+          strictSponsorshipMode: false,
```
AmanRaj1608, line 265: what is fee mode here?
This is updated if user gives in constructor

packages/account-abstraction/src/BiconomyPaymasterAPI.ts
```diff
@@ -42,12 +40,25 @@ export class BiconomyPaymasterAPI implements IPaymasterAPI {
       if (result && result.data && result.statusCode === 200) {
         return result.data.paymasterAndData
       } else {
-        console.error(result.error)
-        throw new Error('Error in verifying. sending paymasterAndData 0x')
+        if (!this.paymasterConfig.strictSponsorshipMode) {
+          return '0x'
+        }
```
AmanRaj1608, line 45: Do we need to throw error here? what gets logged in SDK

livingrockrises: throwing below. currently signing service just throws in all negative cases so it never gets in this block. in future if they fix it we already throw result.error

packages/smart-account/src/SmartAccount.ts
livingrockrises: we can get it from options in constructor. I haven't been able to decide yet would it be at transaction level (sendTransaction unified method with type: and feeMode: ) or at smart account config level (fixed)

AmanRaj1608: I think we should take this in constructor, but will be better if we take in sendTransaction and override maybe

PR #142, opened 2023-04-04
fixes

talhamalik883:
This PR has following changes

1. Networkconfig merge issue in smart account constructor
2. Eip1155 fixes for L2 not having Eip1155
3. Error handling for wallet not deployed case

### DIFF
--- a/packages/common/src/ERC4337Utils.ts
+++ b/packages/common/src/ERC4337Utils.ts
@@ -21,7 +21,7 @@ if (UserOpType == null) {
 
 export const AddressZero = ethers.constants.AddressZero
 
-export const EIP1559_UNSUPPORTED_NETWORKS: Array<ChainId> = [97, 56]
+export const EIP1559_UNSUPPORTED_NETWORKS: Array<ChainId> = [97, 56, 1442, 1101]
 
 // reverse "Deferrable" or "PromiseOrValue" fields
 /* eslint-disable  @typescript-eslint/no-explicit-any */


packages/account-abstraction/src/BaseAccountAPI.ts
```diff
@@ -258,12 +258,20 @@ export abstract class BaseAccountAPI {
...
-      // if wallet is not deployed we need to multiply estimated limit to 3 times to get accurate callGasLimit
+      try {
+        callGasLimit = await this.provider.estimateGas({
+          from: this.entryPoint.address,
+          to: this.getAccountAddress(),
+          data: callData
+        })
+      } catch (error) {
+        Logger.error(' Call Gas Limit Estimation Failed ')
+        if (detailsForUserOp && detailsForUserOp.target.length === 1 &&
```
livingrockrises, line 269: what is this piece of code doing? when it will fall into catch block and only based on this check custom gas price is applied..

packages/account-abstraction/src/SmartAccountAPI.ts
```diff
@@ -190,12 +190,15 @@ export class SmartAccountAPI extends BaseAccountAPI {
     if (maxFeePerGas == null || maxPriorityFeePerGas == null) {
       const feeData = await this.provider.getFeeData()
       if (maxFeePerGas == null) {
-        maxFeePerGas = feeData.maxFeePerGas ?? undefined
+        maxFeePerGas = feeData.maxFeePerGas ?? feeData.gasPrice ?? undefined
```
livingrockrises, line 193: please check my PR also
https://github.com/bcnmy/biconomy-client-sdk/pull/141/files

packages/smart-account/src/SmartAccount.ts
```diff
@@ -154,7 +154,11 @@ class SmartAccount extends EventEmitter {
 
     if (config) {
       const customNetworkConfig: NetworkConfig[] = config.networkConfig || []
-      networkConfig = _.unionBy(customNetworkConfig, networkConfig, 'chainId')
+
+      if ( customNetworkConfig.length !== 0 ){
+        const mergedNetworkConfig = _.merge(_.keyBy(customNetworkConfig, 'chainId'), _.keyBy(networkConfig, 'chainId'))
```
livingrockrises, line 159: need to check this with various network config values

livingrockrises (commented):
reviewed

packages/account-abstraction/src/BaseAccountAPI.ts
talhamalik883: in zkevm polygon case, it falls to catch block, we try to give hardcoded value but only in case its not just wallet deployment transaction. Because for wallet deployment we already have a hardcode number of 500k

livingrockrises: why can't it estimate gas for zkevm?

livingrockrises: also what is the point of this?

if (detailsForUserOp && detailsForUserOp.target.length === 1 && detailsForUserOp.target[0] !== AddressZero)

how does it ensure it's just wallet deployment transaction? if to address is not zero it's wallet deployment, how?

talhamalik883: > why can't it estimate gas for zkevm?

it just throw error saying. Gas estimation failed, unpredictable gas limit

talhamalik883: > also what is the point of this?
> 
> if (detailsForUserOp && detailsForUserOp.target.length === 1 && detailsForUserOp.target[0] !== AddressZero)
> 
> how does it ensure it's just wallet deployment transaction? if to address is not zero it's wallet deployment, how?

Han let say i am just doing wallet deployment using AA then detailsForUserOp.target length would be one and to is going to be address zero other then that we will just use hardcoded gaslimit

livingrockrises: single transaction and to address non-zero -> how does it signify wallet deployment?!

livingrockrises: I would still buy it if was check for 0 address

talhamalik883: > single transaction and to address non-zero -> how does it signify wallet deployment?!

Han to address can be zero for other transaction i will just throw error if we are not able to estimate gas. 

Will post issue in zkevm channel regarding gas estimation

PR #145, opened 2023-04-07
feat: getSmartAccountAddress api added

AmanRaj1608:


packages/smart-account/src/SmartAccount.ts
```diff
@@ -1009,6 +1009,20 @@ class SmartAccount extends EventEmitter {
...
+   * @param address EOA address
+   * @param chainId optional chainId
+   * @param index optional index for counterfactual address
+   * @returns SmartAccount address for given EOA address
+   */
+  getSmartAccountAddress(address: string, chainId?: ChainId, index?: number): Promise<string> {
+    chainId = chainId ? chainId : this.#smartAccountConfig.activeNetworkId
+    index = index ? index : 0
+    const factoryAddr = this.contractUtils.smartWalletFactoryContract[chainId][this.DEFAULT_VERSION]
+    return factoryAddr.getAddressForCounterFactualAccount(address, index)
```
livingrockrises, line 1023: don't you need await here

livingrockrises: I am adding async. I'd also say make it a dto so chainId and index can be truly optional. currently if want to use owner and different index then i still need to give chainId.

livingrockrises (commented):
reviewed

livingrockrises (approved):

PR #157, opened 2023-05-04
fix transactions event emitter

talhamalik883:
transaction events emitter fix


packages/smart-account/src/SmartAccount.ts
```diff
@@ -587,7 +587,7 @@ class SmartAccount extends EventEmitter {
     if (updateImplTrx.data != '0x') {
       transactions.unshift(updateImplTrx)
     }
-    const response = await aaSigner.sendTransactionBatch(transactions, false)
+    const response = await aaSigner.sendTransactionBatch(transactions, this)
```
livingrockrises, line 590: why false argument is changed to 'this'
wouldn't that mess up with the order?

livingrockrises (commented):
reviewed. how sendTransaction and sendTransactionBatch is defined in 4337Signer class?

packages/smart-account/src/SmartAccount.ts
talhamalik883: It was wrong initially, sendTransactionBatch function just take two arguments. One is transaction object and other is event emitter, initially false value was supplied that was wrong so have corrected that

livingrockrises: and sendTransaction takes three args now? without breaking ethers signer interface

talhamalik883: yes, here are the args

```
 transaction: TransactionRequest,
 walletDeployOnly = false,
 engine?: any // EventEmitter
```

livingrockrises:
ok, can you please change the base to development. It won't directly be merged into master..

livingrockrises (approved):
LGTM

AmanRaj1608 (approved):

PR #165, opened 2023-05-10
feat: fetching gas and gas prices from bundler

AmanRaj1608:
Fetch gas and gas price from bundler
`maxPriorityFeePerGas`, `maxFeePerGas` `callGasLimit` `verificationGasLimit` `preVerificationGas`


packages/account-abstraction/src/SmartAccountAPI.ts
```diff
@@ -188,49 +190,40 @@ export class SmartAccountAPI extends BaseAccountAPI {
...
+    }
+
+    let { maxFeePerGas, maxPriorityFeePerGas } = info
+    const chainId = this.clientConfig.chainId
+
+    const feeData = await this.httpRpcClient.getUserOpGasAndGasPrices(partialUserOp)
+    Logger.log('feeData', feeData)
+
+    if (EIP1559_UNSUPPORTED_NETWORKS.includes(chainId)) {
+      maxFeePerGas = undefined
```
talhamalik883, line 210: maxFeePerGas and maxPriorityFeePerGas should not be undefined

you can assign this to both fields

feeData.gasPrice ?? (await this.provider.getGasPrice())

livingrockrises, line 210: if we manage to get feeData from bundler assumption is it would have already considered EIP1559 support and returned valid and same values for maxFeePerGas and maxPriorityFeePerGas

packages/account-abstraction/src/SmartAccountAPI.ts
```diff
@@ -188,49 +190,40 @@ export class SmartAccountAPI extends BaseAccountAPI {
...
+      maxPriorityFeePerGas = undefined
+    } else {
+      // if type 2 transaction, use the gas prices from the user or the default gas price
+      if (!maxFeePerGas)
+        maxFeePerGas = feeData.maxFeePerGas ?? (await this.provider.getGasPrice()) ?? undefined
```
livingrockrises, line 215: since it is type2 it should have been getting maxFeePerGas by making provider.getFeeData call like it is done in the current code in master.

packages/account-abstraction/src/SmartAccountAPI.ts
livingrockrises: yeah he is right, check my comments too

livingrockrises: control flow should be

{ maxFeePerGas, maxPriorityFeePerGas } = info

try feeData from bundler

feeData object -> 

       check maxFeePerGas - yes - use as it is (make sure types is correct)
                                             - no - fallback to provider.getFeeData - check EIP1559 - use accordingly or same values(gasPrice)
                                             
         same for mxPriority                         

catch {

   // previous logic
   
   // get provider.getFeeData // check EIP1559 //assign accordingly

}


cc @talhamalik883

AmanRaj1608:
- It first destructures maxFeePerGas and maxPriorityFeePerGas from info.
- It retrieves the chainId from this.clientConfig and initializes feeData.
- The code then enters a try-catch block to call this.httpRpcClient.getUserOpGasAndGasPrices(partialUserOp). If the call is successful, feeData is updated.
- Inside the try block, if either maxFeePerGas or maxPriorityFeePerGas is not provided by the user (== null), the code checks if the current network is in the EIP1559_UNSUPPORTED_NETWORKS list. If it is, both fees are set to feeData.gasPrice. If not, they are set to the respective fields from feeData.
- In case of an exception, the catch block logs the error and attempts to fetch fee data from this.provider.getFeeData(). Similar to the try block, it checks if the fees were not provided by the user and if the network is in the unsupported list. In this case, the fees are set to feeData.gasPrice or the return of this.provider.getGasPrice(). If the fees are still null, they're set to the respective fields from feeData.
- After the try-catch block, the maxFeePerGas and maxPriorityFeePerGas are assigned to the partialUserOp.
- The preVerificationGas and verificationGasLimit fields of partialUserOp are also set from feeData or by calling this.getPreVerificationGas(partialUserOp) and this.getVerificationGasLimit().

packages/account-abstraction/src/SmartAccountAPI.ts
```diff
@@ -188,49 +190,40 @@ export class SmartAccountAPI extends BaseAccountAPI {
...
+          feeData.maxPriorityFeePerGas ?? (await this.provider.getGasPrice()) ?? undefined
     }
 
-    partialUserOp.paymasterAndData = '0x'
-    const preVerificationGas = await this.getPreVerificationGas(partialUserOp)
-    partialUserOp.preVerificationGas = preVerificationGas
+    partialUserOp.maxFeePerGas = maxFeePerGas
+    partialUserOp.maxPriorityFeePerGas = maxPriorityFeePerGas
+    partialUserOp.preVerificationGas =
+      feeData.preVerificationGas ?? (await this.getPreVerificationGas(partialUserOp))
```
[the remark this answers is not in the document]
AmanRaj1608: It is using the same logic we are using in sdk, can verify on call

livingrockrises (approved):
can go ahead with testing

PR #185, opened 2023-06-09
feat: counterfactual address offchain

AmanRaj1608:
An offchain method to fetch the counterfactual address for the user

The `minimalHandler`, `basicImplementation`, `factoryAddress`, `proxyContractByteCode` can be moved to config file or can be fetched as an parameter.


livingrockrises (approved):
LGTM. this will only work for implementation version V1 
( https://github.com/bcnmy/scw-contracts/tree/prod-v2.0.0 )

livingrockrises:
closing since this package is deprecated for further development

PR #188, opened 2023-06-12
Feat/versioning

talhamalik883:
This PR includes following changes
1. Integration of backend node api call for getting smart account address for EOA
2. Enable versioning
3. bug fix in sendUserOp function. it was calling its self instead of sendSignedUserOp


packages/account/src/BiconomySmartAccount.ts
```diff
@@ -1,20 +1,22 @@
...
+  Logger,
   NODE_CLIENT_URL,
   RPC_PROVIDER_URLS,
-  EntryPoint_v100__factory,
   SmartAccountFactory_v100,
-  SmartAccountFactory_v100__factory,
-  SmartAccount_v100__factory
+  getEntryPointContractInstance,
+  getFactoryContractInstance,
+  getProxyContractInstance
```
tomarsachin2271, line 11: Rename getProxyContractInstance to getSAProxyContract

We can remove Instance from all other names as well.

It should be `getEntryPointContract`, `getSAFactoryContract`, `getSAProxyContract`

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -23,13 +25,15 @@ import {
...
 
 export class BiconomySmartAccount extends SmartAccount implements IBiconomySmartAccount {
-  private factory: SmartAccountFactory_v100
+  private factory!: SmartAccountFactory_v100
   private nodeClient: INodeClient
-  private accountIndex!: Number
+  private accountIndex!: number
   private address!: string
+  private smartAccountInfo!: ISmartAccount
+  private isInited!: boolean
```
tomarsachin2271, line 36: Rename `isInited` to `isInitialized`

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -74,38 +75,89 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
...
+      Logger.error(`Failed to call init: ${error}`);
       throw error
     }
 
     return this
   }
 
+  private isInitialized(): boolean{
+    if (!this.isInited)
+    throw new Error('BiconomySmartAccount is not initialized. Please call BiconomySmartAccount.init() before interacting with any other function')
```
tomarsachin2271, line 92: BiconomySmartAccount.init() in message seems like init() is static function in the class.

We can modify the message like this

..... Please call init() on BiconomySmartAccount instance before interacting ......

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -74,38 +75,89 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
...
+    this.setProxyContractState()
+    this.setEntryPointContractState()
+    this.setFactoryContractState()
+  }
+
   async setAccountIndex(accountIndex: number): Promise<void> {
     this.accountIndex = accountIndex
     this.address = await this.getSmartAccountAddress(accountIndex)
-    this.proxy = await SmartAccount_v100__factory.connect(this.address, this.provider)
+    await this.setContractsState()
```
tomarsachin2271, line 138: Call these methods (setContractState() and setInitCode()) separately in init method... this method name is setAccountIndex and it should only set the account index.

packages/account/src/BaseAccount.ts
```diff
@@ -18,12 +18,21 @@ export abstract class SmartAccount implements ISmartAccount {
...
   signer!: Signer
+  smartAccountConfig: SmartAccountConfig
 
-  constructor(readonly smartAccountConfig: SmartAccountConfig) { }
+  constructor(_smartAccountConfig: SmartAccountConfig) {
+    this.smartAccountConfig = _smartAccountConfig
+   }
+
+
+  setEntrypointAddress(entryPointAddress: string){
```
tomarsachin2271, line 33: Use proper camel case in function names.
It should be `setEntryPointAddress`

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -74,38 +75,89 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
...
-      const address = await this.factory.getAddressForCounterFactualAccount(
-        await this.signer.getAddress(),
-        ethers.BigNumber.from(accountIndex)
-      )
-      return address
+      let smartAccountsList: ISmartAccount[] = (await this.getSmartAccountsByOwner({
+        chainId: this.chainId,
+        owner: this.owner
+      })).data
+      smartAccountsList = smartAccountsList.filter((smartAccount: ISmartAccount) => { return accountIndex === smartAccount.index })
```
tomarsachin2271, line 149: Do proper error handling here. Handle the case if smartAccountList is undefined.

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -172,7 +229,9 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
       initCode: nonce.eq(0) ? this.initCode : '0x',
       callData: callData
     }
+
     userOp = await this.estimateUserOpGas(userOp, overrides)
+    Logger.log('userOp after estimation ', userOp)
```
tomarsachin2271, line 234: These Logger.log across all files should be called based on a debug flag in class constructor.... SDK should not print any logs on its own unless developer has enabled it using debug flag.

packages/account/src/utils/Constants.ts
```diff
@@ -1,14 +1,20 @@
...
+import { EntrypointAddresses, BiconomyFactories, BiconomyImplementation } from './Types'
 
-export const ENTRYPOINT_ADDRESSES = {
-    'V0_0_5': '0x27a4Db290B89AE3373ce4313cBEaE72112Ae7Da9',
-    'V0_0_6': '0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789',
-    default: '0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789'
+export const ENTRYPOINT_ADDRESSES: EntrypointAddresses = {
+    '0x27a4Db290B89AE3373ce4313cBEaE72112Ae7Da9': 'V0_0_5',
+    '0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789': 'V0_0_6',
+    default: '0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789' // will always be latest { ENTRYPOINT_ADDRESSES }
```
tomarsachin2271, line 7: This is not consistent.. for some entries it return the version and for default it returns the address. We should separate these in different data structure.

packages/core-types/src/SmartAccountTypes.ts
```diff
@@ -140,3 +140,9 @@ export type TransactionBatchDto = {
   transactions: Transaction[]
   chainId?: ChainId
 }
+
+export enum SmartAccountType {
+  CANDIDE,
```
tomarsachin2271, line 145: Export CANDIDE and SOUL only when we give support of them in sdk.

packages/account/src/BiconomySmartAccount.ts
talhamalik883: resolved

packages/account/src/BiconomySmartAccount.ts
talhamalik883: resolved

packages/account/src/BaseAccount.ts
talhamalik883: resolved

packages/account/src/BiconomySmartAccount.ts
talhamalik883: there is already a check in place. filter function returns array so undefined case don't need to be managed
if (smartAccountsList.length === 0)
        throw new Error('Failed to get smart account address')

packages/account/src/BiconomySmartAccount.ts
talhamalik883: so logger only logs if debug is enabled. this check is already there in log function

packages/core-types/src/SmartAccountTypes.ts
talhamalik883: resolved

packages/account/src/utils/Constants.ts
talhamalik883: resolved

packages/account/src/BiconomySmartAccount.ts
livingrockrises: can just keep boolean called initialized

livingrockrises:
Is this versioning specific to entry point? 
In PR description it should be clear outcomes about the change

talhamalik883:
> 

we've got versioning all set up for the entrypoint, factory, and implementation. We're using this backend node API called getSmartAccountByOwner to grab all the info about a smart account. This includes the factory address, implementation address, and entry point address that are all linked to the smart account.

Based on the addresses we get from that API call, we're going to decide which version of typechain to use. 
We can interact with the contract using the specific version of typechain that matches with those addresses.

here is how we have managed versions locally

`
export const ENTRYPOINT_ADDRESSES: EntrypointAddresses = {
    '0x27a4db290b89ae3373ce4313cbeae72112ae7da9': 'V0_0_5',
    '0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789': 'V0_0_6'
}
export const BICONOMY_FACTORY_ADDRESSES: BiconomyFactories = {
    '0x000000f9ee1842bb72f6bbdd75e6d3d4e3e9594c': 'V1_0_0'
}
export const BICONOMY_IMPLEMENTATION_ADDRESSES: BiconomyImplementation = {
    '0x00006b7e42e01957da540dc6a8f7c30c4d816af5': 'V1_0_0'
}

`

packages/account/src/BiconomySmartAccount.ts
tomarsachin2271: I'm saying what if smartAccountsList is null

packages/account/src/BiconomySmartAccount.ts
tomarsachin2271: Then rename the method to `initializeAccountAtIndex` to accommodate the actions it's taking.

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -89,41 +89,47 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
...
   private isInitialized(): boolean{
     if (!this.isInited)
-    throw new Error('BiconomySmartAccount is not initialized. Please call BiconomySmartAccount.init() before interacting with any other function')
+    throw new Error('BiconomySmartAccount is not initialized. Please call init() on BiconomySmartAccount before interacting with any other function')
     return true
   }
 
   private setProxyContractState(){
+    if ( !BICONOMY_IMPLEMENTATION_ADDRESSES[this.smartAccountInfo.implementationAddress] )
+    throw new Error('Could not find attach implementation address again your smart account. Please generate support ticket for further investegation.')
```
tomarsachin2271, line 98: Typo in the message, it should be

.. **attached** implementation address **against** your smart account.  
not 
... **attach** implementation address **again** your smart account.

Last message should be
**Please raise an issue on https://github.com/bcnmy/biconomy-client-sdk for further investigation.**

Same goes for other messages.

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -89,41 +89,47 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
 
   private isInitialized(): boolean{
     if (!this.isInited)
-    throw new Error('BiconomySmartAccount is not initialized. Please call BiconomySmartAccount.init() before interacting with any other function')
+    throw new Error('BiconomySmartAccount is not initialized. Please call init() on BiconomySmartAccount before interacting with any other function')
```
tomarsachin2271, line 92: on BiconomySmartAccount **instance**

packages/account/src/BiconomySmartAccount.ts
talhamalik883: resolved

packages/account/src/BiconomySmartAccount.ts
talhamalik883: resolved

packages/account/src/BiconomySmartAccount.ts
talhamalik883: resolved

packages/account/src/BiconomySmartAccount.ts
talhamalik883: That is not going to happen, as the backend API ensures the return of the account address. just in case added this
`
 if ( !smartAccountsList )
      throw new Error('Failed to get smart account address. Please raise an issue on https://github.com/bcnmy/biconomy-client-sdk for further investigation.')
`

packages/account/src/BiconomySmartAccount.ts
tomarsachin2271: **attach** still not changed to **attached**

talhamalik883: resolved

tomarsachin2271 (approved):

packages/account/src/BiconomySmartAccount.ts
tomarsachin2271: From where is the Logger class getting the debug flag?

PR #201, opened 2023-07-03
multi smart account creation and acme dapp support

talhamalik883:
This PR includes:

1. Backend Node API integration that returns smart account details based on EOA and index.
2. Acme dapp support for creating userOps using the signer's address


packages/account/src/BiconomySmartAccount.ts
```diff
@@ -184,7 +202,7 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
     this.initCode = ethers.utils.hexConcat([
       this.factory.address,
       this.factory.interface.encodeFunctionData('deployCounterFactualAccount', [
-        await this.signer.getAddress(),
+        this.owner,
```
livingrockrises, line 205: setInitCode will always be called after setting intended final owner as this.owner state?

packages/node-client/src/NodeClient.ts
```diff
@@ -91,9 +91,11 @@ class NodeClient implements INodeClient {
   async getSmartAccountsByOwner(
     smartAccountByOwnerDto: SmartAccountByOwnerDto
   ): Promise<SmartAccountsResponse> {
-    const { chainId, owner } = smartAccountByOwnerDto
+    const { chainId, owner, index } = smartAccountByOwnerDto
     return sendRequest({
-      url: `${this.#txServiceBaseUrl}/smart-accounts/chainId/${chainId}/owner/${owner}`,
+      url: `${
```
livingrockrises, line 96: where do we set this endpoint as default value?

livingrockrises (approved):
reviewed and left comments. LGTM otherwise

packages/account/src/BiconomySmartAccount.ts
talhamalik883: yeah it is

packages/node-client/src/NodeClient.ts
talhamalik883: sorry don't get what you mean by default value. this is new endpoint with /chainid/owner/index. previous endpoint is also there in backend as /chainid/owner

packages/account/src/BiconomySmartAccount.ts
livingrockrises: ok

packages/node-client/src/NodeClient.ts
livingrockrises: I mean we would be passing different base url for this version of sdk right?

PR #227, opened 2023-07-10
fix userOp estimate gas overrides

talhamalik883:
This PR passing required userOp params for bundler estimation and removal of dummy userOp values

### DIFF
--- a/packages/account/src/BiconomySmartAccount.ts
+++ b/packages/account/src/BiconomySmartAccount.ts
@@ -255,6 +255,14 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
     return executeBatchCallData
   }
 
+  getDummySignature(): string {
+    return '0x73c3ac716c487ca34bb858247b5ccf1dc354fbaabdd089af3b2ac8e78ba85a4959a2d76250325bd67c11771c31fccda87c33ceec17cc0de912690521bb95ffcb1b'
+  }
+
+  getDummyPaymasterData(): string {
+    return '0x'
+  }
+
   async buildUserOp(
     transactions: Transaction[],
     overrides?: Overrides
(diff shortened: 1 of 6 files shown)


packages/account/src/BiconomySmartAccount.ts
```diff
@@ -255,6 +255,14 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
     return executeBatchCallData
   }
 
+  getDummySignature(): string {
+    return '0x73c3ac716c487ca34bb858247b5ccf1dc354fbaabdd089af3b2ac8e78ba85a4959a2d76250325bd67c11771c31fccda87c33ceec17cc0de912690521bb95ffcb1b'
```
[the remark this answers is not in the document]
livingrockrises: if we give param signature it self it just accept param and returns the same?
what else could be input param for this method, can you think in terms of ownerless - I know lot of info there is being passed as part of userop signature cc @filmakarov 

However this class is limited to ECDSA Biconomy Smart Account always so implementing abstract method is correct. If you know what param should be there for this method in abstract class, I can add it.
Anyway there might be more changes to the abstract class also when we implement ownerless. this is the only use case I can think of for your point

AmanRaj1608 (approved):

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -286,6 +294,9 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
       callData: callData
     }
 
+    // for this Smart Account dummy ECDSA signature will be used to estimate gas
+    userOp.signature = this.getDummySignature()
```
AmanRaj1608, line 298: but even if sdk doesn’t send a dummy signature the bundler still appends a dummy signature right? do we need to add this @talhamalik883

livingrockrises: bundler wouldn't know account information. I think it accepts a dummy sig and if not present defaults to ECDSA mock signature.

PR #228, opened 2023-07-11
flag added for choosing userOpGas estimation strategy

talhamalik883:
This PR includes following changes

1. Dto is added for buildUserOp and estimateUserOpGas
2. flag is added in estimateUserOpGasDto to changes estimation logic from bundler to local implementation

### DIFF
--- a/packages/account/src/SmartAccount.ts
+++ b/packages/account/src/SmartAccount.ts
@@ -9,7 +9,7 @@ import { packUserOp } from '@biconomy/common'
 import { IBundler, UserOpResponse } from '@biconomy/bundler'
 import { IPaymaster, PaymasterAndDataResponse } from '@biconomy/paymaster'
 import { EntryPoint_v100, SmartAccount_v100, Logger } from '@biconomy/common'
-import { SmartAccountConfig, Overrides } from './utils/Types'
+import { SmartAccountConfig, EstimateUserOpGasDto } from './utils/Types'
 
 type UserOperationKey = keyof UserOperation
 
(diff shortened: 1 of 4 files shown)


packages/account/src/SmartAccount.ts
```diff
@@ -65,69 +65,50 @@ export abstract class SmartAccount implements ISmartAccount {
...
-          userOp.verificationGasLimit ?? (await this.getVerificationGasLimit(userOp.initCode))
-      userOp.callGasLimit =
-        userOp.callGasLimit ??
-        (await this.provider.estimateGas({
-          from: this.smartAccountConfig.entryPointAddress,
-          to: userOp.sender,
-          data: userOp.callData
-        }))
-      userOp.preVerificationGas = userOp.preVerificationGas ?? this.getPreVerificationGas(userOp)
+      finalUserOp = await this.calculateUserOpGas(userOp)
```
livingrockrises, line 111: pass the original userop to it (which we earlier saved in finalUserOp)
instead of passing the op merged with the overrides.
op merged with overrides should only be used for estimateUserOp gas : bundler call

packages/account/src/utils/Types.ts
```diff
@@ -49,3 +49,15 @@ export type InitilizationData = {
...
   signerAddress?: string
 }
+
+export type EstimateUserOpGasDto = {
+  userOp: Partial<UserOperation>
+  overrides?: Overrides
+  skipBundlerGasEstimation?: boolean
+}
+
+export type BuildUserOpDto = {
```
livingrockrises, line 59: this is tricky because now they will have to pass buildUserOp({transactions: [tx1, tx2, tx3]}

livingrockrises (commented):
reviewed

packages/account/src/SmartAccount.ts
talhamalik883: resolved

packages/account/src/utils/Types.ts
talhamalik883: yes, if we need to go with DTO then its mandatory

PR #238, opened 2023-07-21
verify account deployment even when nonce is 0

talhamalik883:
An RPC call is made to verify whether the smart account is deployed when the nonce is zero. We previously implemented logic to include initCode in userOp only when the nonce is zero. However, we discovered a scenario where someone might have deployed the smart account on your behalf, resulting in a nonce of zero. In this case, the initCode should not be added to userOp.

### DIFF
--- a/packages/account/src/BiconomySmartAccount.ts
+++ b/packages/account/src/BiconomySmartAccount.ts
@@ -288,10 +288,14 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
     } catch (error) {
       // Not throwing this error as nonce would be 0 if this.nonce() throw exception, which is expected flow for undeployed account
     }
+    let isDeployed = true
+
+    isDeployed = nonce.eq(0) ? await this.isAccountDeployed(this.address) : true
+ 
     let userOp: Partial<UserOperation> = {
       sender: this.address,
       nonce,
-      initCode: nonce.eq(0) ? this.initCode : '0x',
+      initCode: !isDeployed ? this.initCode : '0x',
       callData: callData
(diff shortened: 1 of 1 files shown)


livingrockrises:
don't merge this to master. please update the base to development.

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -288,10 +288,14 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
     } catch (error) {
       // Not throwing this error as nonce would be 0 if this.nonce() throw exception, which is expected flow for undeployed account
     }
+    let isDeployed = true
+
+    isDeployed = nonce.eq(0) ? await this.isAccountDeployed(this.address) : true
```
livingrockrises, line 293:
```
if(nonce.eq(0)) {
 isDeployed = await this.isAccountDeployed(this.address)
}
```

above code would be more readable. lmk what you think

livingrockrises (commented):
reviewed

packages/account/src/BiconomySmartAccount.ts
talhamalik883: > ```
> if(nonce.eq(0)) {
>  isDeployed = await this.isAccountDeployed(this.address)
> }
> ```
> 
> above code would be more readable. lmk what you think

yeah, i think yours approach is more readable. I was thinking of multiple cases and used ternary but removed other cases as they were not valid and did not changed the logic from ternary approach to if statement

livingrockrises (approved):
LGTM

PR #249, opened 2023-08-03
(DO NOT MERGE) SmartAccount V2 integration

livingrockrises:
# Description

Adds structure and support for ownerless smart account.
New packages are created : modules, signers

Account implementations 
V1 => latest audited version by C4 and currently deployed in prod and the latest sdk 3.1.1-alpha.0
V2 => ownerless with currently supported validation/auth modules
V3 => This would be 6900 compatible modular SA

I have changed naming conventions below.
SmartAccount -> BaseSmartAccount
ISmartAccount -> IBaseSmartAccount
 
 For this change, I have also carefully reviewed account API (infinitism sdk adds suffix API to smart account classes) structuring and configurations (aka API params) 
 
 For interfaces and types also I have made appropriate hierarchical changes for extending types. BaseAccount interface was missing some must have functions and had some unnecessary functions which do not apply - this has been corrected. BiconomySmartAccount interface should also have minimal necessary methods. 
 
final reviews should search for keywords TODO, Review and Note  : where all the dev notes are added
 
essentially, modules package provide validation module implementation (currently with abstract and child classes -without interfaces) APIs to be used in the SmartAccount class. 
in V2 class, additional params (for account config) to be provided are instance of default validation module (aka the default module used for account deployment) and validation module (if one of the enabled module is to be used for userop dispatching) these two could be the same as default validation module can be used persistently. 

modules define abstract Signer which can be implemented depending on the nature of validation module. hence, signer package is created to export such signers. please read [here](https://github.com/bcnmy/biconomy-client-sdk/blob/livingrock_patch_m_dev_ownerless/packages/signers/src/index.ts) more notes for development plan. 

currently, V2 can use any validation module and ECDSA has been written under modules package. 
Opening room for discussions...
 

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

# How Has This Been Tested?

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas


AmanRaj1608 (commented):
Like the idea of having v1,v2,v3 for backward compatibility support. So the v1 cclients will be interacting with the accounts package class.

What will be the entry point for the v2 users.
If we have the signers package will it have all the methods internally to send transaction or it is just an interface and it will be passed to the account as main package.

What is the purpose of modules package here, will it generate the data for account package and then pass the data to the accounts class?

livingrockrises:
> Like the idea of having v1,v2,v3 for backward compatibility support. So the v1 clients will be interacting with the accounts package class.
> 
> What will be the entry point for the v2 users. If we have the signers package will it have all the methods internally to send transaction or it is just an interface and it will be passed to the account as main package.
> 
> What is the purpose of modules package here, will it generate the data for account package and then pass the data to the accounts class?

 entry point for the v2 users is also account package -> import BiconomySmartAccountV2 class and provide appropriate config including the module instance. 
 
 signer currently doesn't have internally method to send transaction. we can later add common BiconomySASigner and BiconomySAProvider classes which abstracts accordingly. currently signer is intended to be passed in module instance depending on the nature of the module required signer can be passed, which helps with methods like getAddress signMessage etc.

PR #258, opened 2023-08-20
Session Key Manager module implementation

AmanRaj1608:
SessionKeyManagerModule implementation

#### User flow to add a session key

```ts
const biconomyAccount = new BiconomySmartAccount();
const biconomySessionManager = await SessionKeyManagerModule.create();

const sessionKeyData = await createSessionData({
  validUntil: 0,
  validAfter: 0,
  sessionValidationModuleAddress: "0xeeee address of module (eg: erc20)",
  sessionKeyData: "0xabc data",
});

const tx = {
  to: await biconomyAccount.address,
  data: sessionKeyData,
  value: 0,
}

const partialUserOp = await biconomySmartAccount.buildUserOp([tx]);
const userOpResponse = await biconomySmartAccount.sendUserOp(partialUserOp);

// after success tx set the sessionKey in the session manager
if(userOpResponse.status === "success") {
  await biconomySessionManager.setSessionKey(sessionKey);  
}
```

#### User flow to do a transaction with session key

```ts
const biconomyAccount = new BiconomySmartAccount();
const biconomySessionManager = await SessionKeyManagerModule.create();

const tx = {
  to: "0xabc",
  data: "0x",
  value: 0.1,
}

// Not sure what is the process here cc: @livingrockrises @tomarsachin2271 
// - does the SDK calls the signUserOp with the session key?
// - or the user calls the signUserOp with the session key?

const partialUserOp = await biconomySmartAccount.buildUserOp([tx]);
const sig = await biconomySessionManager.signUserOp(partialUserOp, sessionSigner);
partialUserOp.signature = sig;

const userOpResponse = await biconomySmartAccount.sendUserOp(partialUserOp);
```

### DIFF
--- a/packages/modules/src/BaseValidationModule.ts
+++ b/packages/modules/src/BaseValidationModule.ts
@@ -1,8 +1,7 @@
 import { Signer } from 'ethers'
-import { Logger } from '@biconomy/common'
-import { Bytes, BytesLike, hexConcat, hexZeroPad, hexlify } from 'ethers/lib/utils'
+import { Bytes } from 'ethers/lib/utils'
 import { BaseValidationModuleConfig } from './utils/Types'
-import { UserOperation, ChainId } from '@biconomy/core-types'
+import { UserOperation } from '@biconomy/core-types'
 import { DEFAULT_ENTRYPOINT_ADDRESS } from './utils/Constants'
 import { IValidationModule } from './interfaces/IValidationModule'
 
@@ -28,7 +27,7 @@ export abstract class BaseValidationModule implements IValidationModule {
   // Review naming convention for getter
   abstract getSigner(): Promise<Signer>
 
-  abstract signUserOp(userOperation: UserOperation): Promise<string>
+  abstract signUserOp(userOperation: UserOperation, sessionSigner?: Signer): Promise<string>
 
   abstract signMessage(message: Bytes | string): Promise<string>
 }
--- a/packages/modules/src/BaseValidationModule.ts
+++ b/packages/modules/src/BaseValidationModule.ts
@@ -27,7 +27,7 @@ export abstract class BaseValidationModule implements IValidationModule {
   // Review naming convention for getter
   abstract getSigner(): Promise<Signer>
 
-  abstract signUserOp(userOperation: UserOperation, sessionSigner?: Signer): Promise<string>
+  abstract signUserOp(userOperation: UserOperation, signer?: Signer): Promise<string>
 
   abstract signMessage(message: Bytes | string): Promise<string>
 }
--- a/packages/modules/src/SessionKeyManagerModule.ts
+++ b/packages/modules/src/SessionKeyManagerModule.ts
@@ -104,7 +104,7 @@ export class SessionKeyManagerModule extends BaseValidationModule {
     }
 
     await this.sessionStorageClient.addSessionData(sessionLeafNode)
-    // create a signer if sessionPubKey if not given
+    // TODO: create a signer if sessionPubKey if not given
     return setMerkleRootData
   }
 
@@ -126,8 +126,7 @@ export class SessionKeyManagerModule extends BaseValidationModule {
       sessionPublicKey: await sessionSigner.getAddress()
     })
 
-    // Generate the padded signature with
-    // (validUntil, validAfter, sessionVerificationModuleAddress, validationData, merkleProof, signature)
+    // Generate the padded signature with (validUntil,validAfter,sessionVerificationModuleAddress,validationData,merkleProof,signature)
     const paddedSignature = defaultAbiCoder.encode(
       ['uint48', 'uint48', 'address', 'bytes', 'bytes32[]', 'bytes'],
       [
@@ -158,7 +157,13 @@ export class SessionKeyManagerModule extends BaseValidationModule {
     this.sessionStorageClient.updateSessionStatus(param, status)
   }
 
-  // clear all pending sessions
+  /**
+   * @remarks This method is used to clear all the pending sessions
+   * @returns
+   */
+  async clearPendingSessions() {
+    this.clearPendingSessions()
+  }
 
   /**
    * @returns SessionKeyManagerModule address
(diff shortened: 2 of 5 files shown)


packages/modules/src/BaseValidationModule.ts
```diff
@@ -28,7 +27,7 @@ export abstract class BaseValidationModule implements IValidationModule {
   // Review naming convention for getter
   abstract getSigner(): Promise<Signer>
 
-  abstract signUserOp(userOperation: UserOperation): Promise<string>
+  abstract signUserOp(userOperation: UserOperation, sessionSigner?: Signer): Promise<string>
```
livingrockrises, line 30: can we just name this a signer in abstract class. makes more sense

livingrockrises: or you can just omit it. if in child class name to be kept as sessionSigner only.
cc @tomarsachin2271

packages/modules/src/SessionKeyManagerModule.ts
```diff
@@ -1,139 +1,192 @@
...
-    const validAfter = 0
-    const sessionEOA = await this.sessionSigner.getAddress()
-    const sessionKeyData = hexZeroPad(sessionEOA, 20)
-    const newLeafData = hexConcat([
-      hexZeroPad(ethers.utils.hexlify(validUntil), 6),
-      hexZeroPad(ethers.utils.hexlify(validAfter), 6),
-      hexZeroPad(this.getAddress(), 20), // TODO // actually session validation module address
-      sessionKeyData
-    ])
+  // TODO: why this is needed?
```
livingrockrises, line 174: answered this on another PR.

packages/modules/src/SessionKeyManagerModule.ts
```diff
@@ -1,139 +1,203 @@
...
   }
 
-  async createSession(): Promise<string> {
+  /**
+   * Method to create session data for any module. The session data is used to create a leaf in the merkle tree
+   * @param leafData The data to be used to create session data
+   * @returns The session data
+   */
+  createSessionData = async (leafData: CreateSessionDataParams): Promise<string> => {
     const sessionKeyModuleAbi = 'function setMerkleRoot(bytes32 _merkleRoot)'
```
tomarsachin2271, line 88: Rename `sessionKeyModuleAbi` to `sessionKeyManagerModuleABI`

packages/modules/src/SessionKeyManagerModule.ts
```diff
@@ -1,139 +1,203 @@
...
     const sessionKeyModuleInterface = new ethers.utils.Interface([sessionKeyModuleAbi])
```
tomarsachin2271, line 89: Rename to `sessionKeyManagerModuleInterface`

packages/modules/src/SessionKeyManagerModule.ts
```diff
@@ -1,139 +1,203 @@
...
-    return this.moduleAddress
-  }
-
-  async getSigner(): Promise<Signer> {
-    throw new Error('Method not implemented.')
-  }
+    instance.sessionStorageClient = new SessionLocalStorage(moduleConfig.smartAccountAddress)
+
+    const existingSessionData = await instance.sessionStorageClient.getAllSessionData()
+    const existingSessionDataLeafs = existingSessionData.map((sessionData) => {
```
tomarsachin2271, line 64: Exclude the sessions whose validUltil date has passed or status is EXPIRED.

packages/modules/src/SessionKeyManagerModule.ts
```diff
@@ -1,139 +1,203 @@
...
+      throw new Error('Session signer is not provided.')
+    }
+    // Use the sessionSigner to sign the user operation
     const userOpHash = getUserOpHash(userOp, this.entryPointAddress, this.chainId)
-    //
-    const signature = await this.sessionSigner.signMessage(arrayify(userOpHash))
-    // add validator module address to the signature
+    const signature = await sessionSigner.signMessage(arrayify(userOpHash))
+
+    const sessionSignerData = await this.sessionStorageClient.getSessionData({
```
tomarsachin2271, line 125: Search param should also include sessionValidationModule address along with sessionPublicKey. Both have to be present to uniqely find the sessionKeyData.

Or it should search using sessionID only.

So instead of sessionSigner as second param, we should take an object which has
sessionSigner,
sessionValidationModule,
sessionID
additionalSessionData : string

parameters where sessionValidationModule, additionalSessionData and sessionID are optional. But if sessionID is absent, sessionValidationModule address must be present. else if sessionID is present, it is given priority.


additionalSessionData will be optional data that we'll just append in the signature field of the whole padded data.

packages/modules/src/SessionKeyManagerModule.ts
```diff
@@ -1,139 +1,203 @@
...
-    //   this.sessionKeyModule,
-    //   this.sessionKey.getAddress()
-    // )
-    // console.log(merkleProofData)
+  /**
+   * @remarks This method is used to clear all the pending sessions
+   * @returns
+   */
+  async clearPendingSessions() {
+    this.clearPendingSessions()
```
tomarsachin2271, line 165: This should be `this.sessionStorageClient.clearPendingStatus()` otherwise this'll go in recursion forever

packages/modules/src/SessionKeyManagerModule.ts
AmanRaj1608: I believe the merkle tree will still have the data its just it will be validate on chain? It should have all the data as per the discussion we had today

tomarsachin2271: Yeah as per the discussion, all the pending state sessions should be part of the merkle tree. Even though this recommended change doesn't affect the functionality in any way but it's just optimised way of doing things, where we can get rid of expired session leafs. 
Coz over time the leaves will keep on getting added and there's a limit of data that can be stored in local storage.

Not critical but we can add a TODO here to handle this later may be.

PR #262, opened 2023-08-24
fix: incorrect data merkleRoot length

AmanRaj1608:
Fixes the issue reported by rage trade. `ethers/lib/utils` keccak256 returns the 32 bytes output on getHexRoot.

> maybe affected by https://github.com/merkletreejs/merkletreejs/issues/86 or same as https://github.com/merkletreejs/merkletreejs/issues/33 ?

### DIFF
--- a/packages/modules/src/SessionKeyManagerModule.ts
+++ b/packages/modules/src/SessionKeyManagerModule.ts
@@ -1,7 +1,8 @@
 import { Signer, ethers } from 'ethers'
 import MerkleTree from 'merkletreejs'
 import { getUserOpHash, NODE_CLIENT_URL } from '@biconomy/common'
-import { hexConcat, arrayify, keccak256, hexZeroPad, defaultAbiCoder } from 'ethers/lib/utils'
+import { hexConcat, arrayify, hexZeroPad, defaultAbiCoder } from 'ethers/lib/utils'
+import { keccak256 } from 'ethereumjs-util'
 import {
   SessionKeyManagerModuleConfig,
   ModuleVersion,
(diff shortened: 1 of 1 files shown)


livingrockrises (approved):
LGTM

packages/modules/src/SessionKeyManagerModule.ts
```diff
@@ -1,7 +1,8 @@
 import { Signer, ethers } from 'ethers'
 import MerkleTree from 'merkletreejs'
 import { getUserOpHash, NODE_CLIENT_URL } from '@biconomy/common'
-import { hexConcat, arrayify, keccak256, hexZeroPad, defaultAbiCoder } from 'ethers/lib/utils'
+import { hexConcat, arrayify, hexZeroPad, defaultAbiCoder } from 'ethers/lib/utils'
+import { keccak256 } from 'ethereumjs-util'
```
livingrockrises, line 5: do we need to use keccak256 from ethereumjs-util in other places too?

AmanRaj1608: Only for generating the merkle root this is required for now as the `addLeaves` needs buffer which we was generating and the length was incorrect

livingrockrises: in MultiChain Validator it would be fine as we use keccak256 as arg in MerkleTree class.

```
     import {
  keccak256,
  arrayify,
  defaultAbiCoder,
  hexConcat,
  hexZeroPad,
  Bytes
} from 'ethers/lib/utils'
      

     const merkleTree = new MerkleTree(leaves, keccak256, { sortPairs: true })

      const multichainSignature = await this.signer.signMessage(arrayify(merkleTree.getHexRoot()))

      Logger.log('merkle root ', merkleTree.getHexRoot())
```

PR #265, opened 2023-08-25
Set merkle root in session storage while creating session

tomarsachin2271:
Adding merkle root in session storage when a new session is created and stored in the storage.


livingrockrises (approved):
LGTM

PR #275, opened 2023-09-01
added session router + updated existing session manager module + dev …

livingrockrises:
…notes

# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

# How Has This Been Tested?

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas


packages/modules/src/BatchedSessionRouterModule.ts
```diff
@@ -0,0 +1,290 @@
...
+    if (moduleConfig.moduleAddress) {
+      instance.moduleAddress = moduleConfig.moduleAddress
+    } else if (moduleConfig.version) {
+      const moduleAddr = BATCHED_SESSION_ROUTER_MODULE_ADDRESSES_BY_VERSION[moduleConfig.version]
+      if (!moduleAddr) {
+        // throw new Error(`Invalid version ${moduleConfig.version}`)
+        Logger.error(`Invalid version ${moduleConfig.version}`)
+        instance.moduleAddress = DEFAULT_BATCHED_SESSION_ROUTER_MODULE
+      }
+      instance.moduleAddress = moduleAddr
```
tomarsachin2271, line 62: Should throw inside the if statement. Otherwise `instance.moduleAddress` will be initialised as undefined in the next statement after if block.

packages/modules/src/BatchedSessionRouterModule.ts
```diff
@@ -0,0 +1,290 @@
...
+      ])
+      return ethers.utils.keccak256(leafDataHex)
+    })
+
+    instance.merkleTree = new MerkleTree(existingSessionDataLeafs, keccak256, {
+      sortPairs: true,
+      hashLeaves: false
+    })
+
+    return instance
```
tomarsachin2271, line 94: Let SessionKeyManager module handle the merkle tree state. Like the onchain module, this module also can be made to not store any state related to sessions and just route the request to SessionKeyManager module.

In create method, lets take an instance of SessionKeyManager module to enable this.

Coz in current state, if there's any change in merkle tree logic or storage of sessions, we'll need to change the same logic at multiple places. If you are making it easy for developers to use, then in that case we should create an instance of SessionKeyManager module inside this class and use that instance instead of repeating the same logic here.

packages/modules/src/BatchedSessionRouterModule.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  /**
+   * Method to create session data for any module. The session data is used to create a leaf in the merkle tree
+   * @param leavesData The data of one or more leaves to be used to create session data
+   * @returns The session data
+   */
+  createSessionData = async (leavesData: CreateSessionDataParams[]): Promise<string> => {
+    const sessionKeyManagerModuleABI = 'function setMerkleRoot(bytes32 _merkleRoot)'
+    const sessionKeyManagerModuleInterface = new ethers.utils.Interface([
```
tomarsachin2271, line 104: Let this call sessionKeyManager module's createSessionData method.

packages/modules/src/SessionKeyManagerModule.ts
```diff
@@ -49,7 +52,9 @@ export class SessionKeyManagerModule extends BaseValidationModule {
     } else if (moduleConfig.version) {
       const moduleAddr = SESSION_MANAGER_MODULE_ADDRESSES_BY_VERSION[moduleConfig.version]
       if (!moduleAddr) {
-        throw new Error(`Invalid version ${moduleConfig.version}`)
+        // throw new Error(`Invalid version ${moduleConfig.version}`)
+        Logger.error(`Invalid version ${moduleConfig.version}`)
```
tomarsachin2271, line 56: Should keep throwing the error in case invalid version is passed.

packages/modules/src/SessionKeyManagerModule.ts
```diff
@@ -131,7 +145,11 @@ export class SessionKeyManagerModule extends BaseValidationModule {
    * @param sessionSigner The signer to be used to sign the user operation
    * @returns The signature of the user operation
    */
-  async signUserOpHash(userOpHash: string, sessionParam?: SessionParams): Promise<string> {
+  async signUserOpHash(userOpHash: string, sessionParams?: SessionParams[]): Promise<string> {
+    if (!sessionParams || sessionParams.length === 0 || sessionParams.length > 1) {
+      throw new Error('Session parameters are not provided or invalid')
```
tomarsachin2271, line 150: Error message can be more clear in terms of requirements 

`Session parameters are not provided or of invalid length. Only one object is allowed in sessionParams array.`

packages/modules/src/utils/Constants.ts
```diff
@@ -21,11 +21,19 @@ export const ECDSA_OWNERSHIP_MODULE_ADDRESSES_BY_VERSION = {
   // and so on
 }
 
-export const DEFAULT_SESSION_KEY_MANAGER_MODULE = '0x000000456b395c4e107e0302553B90D1eF4a32e9'
+export const DEFAULT_SESSION_KEY_MANAGER_MODULE = '0x000002FbFfedd9B33F4E7156F2DE8D48945E7489'
 
 export const SESSION_MANAGER_MODULE_ADDRESSES_BY_VERSION = {
   V1_0_0: '0x000000456b395c4e107e0302553B90D1eF4a32e9'
-  // 'V1_0_1' : '0x000000456b395c4e107e0302553B90D1eF4a32e9'
+  // 'V1_0_1' : '0x000002FbFfedd9B33F4E7156F2DE8D48945E7489'
```
tomarsachin2271, line 28: Should uncomment this version. Default module address should also be a part of module list.

packages/modules/src/BatchedSessionRouterModule.ts
AmanRaj1608: I feel we could remove the `merkleTree` instance all together. We always fetch the local storage and update and create new tree instance.

It is only being used in `signUserOpHash`, so if we are okay to fetch the data again in that method we can remove this.

AmanRaj1608 (commented):
whats the dff between the session router module and session manager 

can see similar methods in both

tomarsachin2271:
> 

You can imagine session router being another module from SA point of view. Where SA validateUserOp method will call SessionRouter's validateUserOp and in this module we are expecting different format of userOp.signature, now we expect array of information about different sessions in one signature, all signed by same session key.
This session router is just routing different sessions to SessionKeyManager(SKM) and then SessionValidationModules (SVM)

Common things that's done here is signature verification and then asking SKM , if this session is valid or not (part of merkle tree) and then calling SVMs to see the permissions are correct or not, then extracting common validUntil and validAfter from all session information and returning it back to SA (Smart Account)

You can see the sequence diagram here for better clarity
https://linear.app/biconomy/issue/SMA-110#comment-f6f3eff8

packages/modules/src/utils/Constants.ts
livingrockrises: "Default module address should also be a part of module list." didnt get this
should I not keep default address as another constant above?

packages/modules/src/SessionKeyManagerModule.ts
livingrockrises: ok

packages/modules/src/SessionKeyManagerModule.ts
livingrockrises: even if version is invalid I was thinking to use default module address then which comes from the constants.

packages/modules/src/BatchedSessionRouterModule.ts
livingrockrises: ok, we can create an instance of it within this class as suggested in other comment.

packages/modules/src/BatchedSessionRouterModule.ts
livingrockrises: when generating proof we will need this instance. are you suggesting to fetch everything from localstorage create a tree and get proof?

packages/modules/src/BatchedSessionRouterModule.ts
livingrockrises: ok let me check this

livingrockrises: moved it in else block

packages/modules/src/BatchedSessionRouterModule.ts
livingrockrises: create an instance of SessionKeyManager module inside this class and use that instance - for us to create an instance I will also need SA address

packages/modules/src/BatchedSessionRouterModule.ts
livingrockrises: done

packages/modules/src/SessionKeyManagerModule.ts
livingrockrises: fixing by keeping it in the else part. do review again for final

packages/modules/src/BatchedSessionRouterModule.ts
```diff
@@ -0,0 +1,239 @@
...
+    } else if (moduleConfig.version) {
+      const moduleAddr = BATCHED_SESSION_ROUTER_MODULE_ADDRESSES_BY_VERSION[moduleConfig.version]
+      if (!moduleAddr) {
+        // throw new Error(`Invalid version ${moduleConfig.version}`)
+        Logger.error(`Invalid version ${moduleConfig.version}`)
+        instance.moduleAddress = DEFAULT_BATCHED_SESSION_ROUTER_MODULE
+      } else {
+        instance.moduleAddress = moduleAddr
+      }
+      instance.version = moduleConfig.version as ModuleVersion
```
livingrockrises, line 61: need to review this assignment in case version is invalid and we used default address..

packages/modules/src/utils/Types.ts
```diff
@@ -16,23 +17,40 @@ export interface ECDSAOwnershipValidationModuleConfig extends BaseValidationModu
 export interface SessionKeyManagerModuleConfig extends BaseValidationModuleConfig {
   moduleAddress?: string
   version?: ModuleVersion
-  sessionSigner?: Signer
-  sessionPubKey?: string
+  // Review
+  // sessionSigner?: Signer
```
livingrockrises, line 21: this was only meant to be there if we have global public key for all kinds of session leafs generated for this SA+module

packages/modules/src/utils/Constants.ts
tomarsachin2271: Yes we should keep the constant variable but that value should also be reflected in the versions list as well.

packages/modules/src/BatchedSessionRouterModule.ts
livingrockrises: ok updated like this 

```
if (moduleConfig.moduleAddress) {
      instance.moduleAddress = moduleConfig.moduleAddress
    } else if (moduleConfig.version) {
      const moduleAddr = BATCHED_SESSION_ROUTER_MODULE_ADDRESSES_BY_VERSION[moduleConfig.version]
      if (!moduleAddr) {
        throw new Error(`Invalid version ${moduleConfig.version}`)
      }
      instance.moduleAddress = moduleAddr
      instance.version = moduleConfig.version as ModuleVersion
    } else {
      instance.moduleAddress = DEFAULT_BATCHED_SESSION_ROUTER_MODULE
      // Note: in this case Version remains the default one
    }
```

packages/modules/src/BaseValidationModule.ts
```diff
@@ -29,8 +29,8 @@ export abstract class BaseValidationModule implements IValidationModule {
   // Review naming convention for getter
   abstract getSigner(): Promise<Signer>
 
-  // Review
-  abstract signUserOpHash(userOpHash: string, moduleSignerInfo?: SessionParams[]): Promise<string>
+  // Signer specific or any other additional information can be passed a params
```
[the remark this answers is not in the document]
livingrockrises: it's already 

```
// Signer specific or any other additional information can be passed a params
  abstract signUserOpHash(userOpHash: string, params?: ModuleInfo): Promise<string>
```

packages/modules/src/utils/Constants.ts
livingrockrises: you mean instead of

V1_0_0: '0x0000001c5b32F37F5beA87BDD5374eB2aC54eA8e'

doing this

export const DEFAULT_ECDSA_OWNERSHIP_MODULE = '0x0000001c5b32F37F5beA87BDD5374eB2aC54eA8e'

V1_0_0: DEFAULT_ECDSA_OWNERSHIP_MODULE

packages/modules/src/SessionKeyManagerModule.ts
livingrockrises: done

packages/modules/src/BatchedSessionRouterModule.ts
```diff
@@ -100,7 +100,9 @@ export class BatchedSessionRouterModule extends BaseValidationModule {
   async signUserOpHash(userOpHash: string, params?: ModuleInfo): Promise<string> {
     const sessionParams = params?.batchSessionParams
     if (!sessionParams || sessionParams.length === 0) {
-      throw new Error('Session parameters are not provided.')
+      throw new Error(
+        'Session parameters are not provided or of invalid length. Only one object is allowed in sessionParams array.'
```
tomarsachin2271, line 104: In case of BatchedSessionRouter module, this error message doesn't apply here.

Only one object is allowed in sessionParams array

Now we have batchSessionParams which is expected to be just non-empty.

packages/modules/src/BaseValidationModule.ts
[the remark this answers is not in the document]
tomarsachin2271: I'm saying there's a typo in the message..

Its "... passed a params", it should be ".. passed **as** params"

packages/modules/src/utils/Constants.ts
livingrockrises: this was already there!

packages/modules/src/BaseValidationModule.ts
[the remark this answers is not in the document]
livingrockrises: ok ok

packages/modules/src/BatchedSessionRouterModule.ts
livingrockrises: ok i see! done

PR #281, opened 2023-09-08
SMA-44 Add ERC20 implementation and interface for SessionValidationModule

tomarsachin2271:
# Description

Added Implementation of ERC20SessionValidationModule and interface ISessionValidationModule which can be used by future session validation modules.

## Type of change
- [x] This change requires a documentation update

# How Has This Been Tested?
Will write test cases once testing framework is enabled.

# Checklist:


packages/modules/src/session-validation-modules/ERC20SessionValidationModule.ts
```diff
@@ -0,0 +1,69 @@
+import { defaultAbiCoder } from 'ethers/lib/utils'
+import { ISessionValidationModule } from 'interfaces/ISessionValidationModule'
+import { ERC20SessionKeyData, SessionValidationModuleConfig } from 'utils/Types'
```
livingrockrises, line 3: SessionValidationModuleConfig this is base config?

livingrockrises (approved):

AmanRaj1608 (approved):

packages/modules/src/session-validation-modules/ERC20SessionValidationModule.ts
tomarsachin2271: no.. Base config for normal validation modules is different. For session validation modules, each config would be different and specific to SVM.. Couldn't find any common config that we can keep as base for SVM. So kept it separate.

PR #288, opened 2023-09-12
feat: core package

AmanRaj1608:
A wrapper signer over the biconomy packages to make dev experience better.

Users can do transactions like

```ts
import { ethers } from "ethers";
import { BiconomyProvider } from "@biconomy/core";
import { ECDSAOwnershipValidationModule } from "@biconomy/modules";

export const mintNft = async () => {
  // get EOA address from wallet provider
  let provider = new ethers.providers.JsonRpcProvider("https://rpc-mumbai.maticvigil.com");
  let signer = new ethers.Wallet("", provider); // update pkey here

  const module = new ECDSAOwnershipValidationModule({
    signer: signer,
  });

  const aaProvider = await BiconomyProvider.create({
    chainId: "80001",
    bundlerUrl: "https://bundler.biconomy.io/api/v2/80001/A5CBjLqSc.0dcbc53e-anPe-44c7-b22d-21071345f76a",
    defaultValidationModule: module,
  });

  // generate mintNft data, minting NFT to smart account address itself
  const nftInterface = new ethers.utils.Interface([
    "function safeMint(address _to)",
  ]);
  const data = nftInterface.encodeFunctionData("safeMint", [
    await aaProvider.getAddress(),
  ]);
  const nftAddress = "0x1758f42Af7026fBbB559Dc60EcE0De3ef81f665e"; // Todo // use from config
  const transaction = {
    to: nftAddress,
    data: data,
  };

  const tx = await aaProvider.sendTransaction([transaction]);
  const { receipt } = await tx.wait();
  console.log(receipt.transactionHash);
};

mintNft();
```


packages/core/src/BiconomyProvider.ts
```diff
@@ -0,0 +1,157 @@
...
+import { BiconomySmartAccountV2, DEFAULT_ENTRYPOINT_ADDRESS, BuildUserOpOptions } from "@biconomy/account";
+import { BiconomyPaymaster, PaymasterMode } from "@biconomy/paymaster";
+import { IBiconomyProvider } from "./interfaces/IBiconomyProvider";
+import { PaymasterDto, SmartAccountProviderInternalOpts, SmartAccountProviderOpts } from "./Types";
+
+/**
+ * @title BiconomyProvider Class
+ * @dev Implements the IBiconomyProvider interface, acting as a wrapper over various Biconomy packages.
+ */
+export class BiconomyProvider implements IBiconomyProvider {
```
livingrockrises, line 13: should it not extend from BaseProvider? we may be missing some provider properties

packages/core/src/interfaces/IBiconomyProvider.ts
```diff
@@ -0,0 +1,6 @@
+export interface IBiconomyProvider {
+  // TODO: fill in the rest of the interface
```
livingrockrises, line 2: check this todo

livingrockrises:
maybe we can write test cases as well for this approach?
like setting another module as active validation module etc. which I assume sets module on internally created SA API class instance?

PR #294, opened 2023-09-15
fix: gitInitCode cache issue

AmanRaj1608:
The BaseSmartAccount has isDeployed variable in cache, and it stores it on constructor init as default false then read.

It is never updated, so either can updated but in other methods but then will have to add logic if tx is failed or not.

Best way is to remove this check and call rpc to check if deployed or not everytime


livingrockrises (approved):
LGTM.
I updated the test cases to ensure this is working.
Also, we should do overall latency check  (time for SDK to build userop - time taken for API calls) and work on optimisations.

PR #304, opened 2023-09-27
add simulation type param

AmanRaj1608:
If someone wants to have this call data check on their end for every transaction we have enabled more parameters in our eth_sendUserOperation to pass in a simulation_type an example curl for this: https://gist.github.com/arcticfloyd1984/26efb6099efc651d2cf064340c67c6db.

### DIFF
--- a/packages/account/src/utils/Types.ts
+++ b/packages/account/src/utils/Types.ts
@@ -87,11 +87,13 @@ export type NonceOptions = {
   nonceOverride?: number;
 };
 
+// Used in AccountV1
 export type SendUserOpDto = {
   signer?: Signer;
   simulationType?: SimulationType;
 };
 
+// Generic options in AccountV2
 export type SendUserOpOptions = {
   simulationType?: SimulationType;
 };
--- a/packages/account/src/utils/Types.ts
+++ b/packages/account/src/utils/Types.ts
@@ -87,6 +87,13 @@ export type NonceOptions = {
   nonceOverride?: number;
 };
 
+export type SendUserOpOptions = {
+  signer?: Signer;
+  simulationType?: SimulationType;
+};
+
+export type SimulationType = "validation" | "validation_and_execution";
+
 export type Overrides = {
   callGasLimit?: BigNumberish;
   verificationGasLimit?: BigNumberish;
--- a/packages/account/src/utils/Types.ts
+++ b/packages/account/src/utils/Types.ts
@@ -87,8 +87,12 @@ export type NonceOptions = {
   nonceOverride?: number;
 };
 
-export type SendUserOpOptions = {
+export type SendUserOpDto = {
   signer?: Signer;
+  simulationType?: SimulationType;
+};
+
+export type SendUserOpOptions = {
   simulationType?: SimulationType;
 };
 
--- a/packages/bundler/src/interfaces/IBundler.ts
+++ b/packages/bundler/src/interfaces/IBundler.ts
@@ -3,7 +3,7 @@ import { UserOperation } from "@biconomy/core-types";
 
 export interface IBundler {
   estimateUserOpGas(_userOp: Partial<UserOperation>): Promise<UserOpGasResponse>;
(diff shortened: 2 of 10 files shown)


packages/account/src/utils/Types.ts
```diff
@@ -87,6 +87,13 @@ export type NonceOptions = {
   nonceOverride?: number;
 };
 
+export type SendUserOpOptions = {
+  signer?: Signer;
+  simulationType?: SimulationType;
+};
+
+export type SimulationType = "validation" | "validation_and_execution";
```
livingrockrises, line 95: can we make this proper enums? and options as all uppercase

packages/modules/src/utils/Types.ts
```diff
@@ -59,6 +59,12 @@ export type ModuleInfo = {
   batchSessionParams?: SessionParams[];
 };
 
+export interface SendUserOpParams extends ModuleInfo {
+  simulationType: SimulationType;
```
livingrockrises, line 63: why this is not optional here? but for bundler utils Types it is

livingrockrises (commented):
reviewed

packages/bundler/src/interfaces/IBundler.ts
```diff
@@ -1,9 +1,9 @@
-import { UserOpResponse, UserOpGasResponse, UserOpReceipt, UserOpByHashResponse } from "../utils/Types";
+import { UserOpResponse, UserOpGasResponse, UserOpReceipt, UserOpByHashResponse, SendUserOpOptions } from "../utils/Types";
 import { UserOperation } from "@biconomy/core-types";
 
 export interface IBundler {
   estimateUserOpGas(_userOp: Partial<UserOperation>): Promise<UserOpGasResponse>;
-  sendUserOp(_userOp: UserOperation): Promise<UserOpResponse>;
+  sendUserOp(_userOp: UserOperation, _params?: SendUserOpOptions): Promise<UserOpResponse>;
```
livingrockrises, line 6: some places it is params some places it is _params. keep one

also try not to add anything in interfaces and abstract contracts (if their type is not used directly. like SA we use exact class type, for bundler also if its Bundler and ot IBundler can only have this in child and not interfaces)

livingrockrises (commented):
left more comments

packages/bundler/src/interfaces/IBundler.ts
AmanRaj1608: the issue is the account package variable `bundler` type is not the Bundler, it's `IBundler` so if the method is not mentioned there it will throw errror. Or I have to update the account package from 
image

Regarding the `_params`. Have to update all the unused variables with _ before to avoid lint warning, majorly in interface for all variables.

livingrockrises: ok, bundler is fine but in BaseSmartAccount it can be avoided to add optional params and just add in the SA classes. 

and _params case try to keep it consistent then, at least it must be same way in particular file.

AmanRaj1608: > and _params case try to keep it consistent then, at least it must be same way in particular file.

updated

> ok, bundler is fine but in BaseSmartAccount it can be avoided to add optional params and just add in the SA classes.

I don't think this is possible because the SA class is not calling the bundler at all, the BaseSmartAccount class calls the bundler and it should have the context to that file.

AmanRaj1608: > and _params case try to keep it consistent then, at least it must be same way in particular file.

updated

> ok, bundler is fine but in BaseSmartAccount it can be avoided to add optional params and just add in the SA classes.

This will work for the SA class but I don't think this is possible because the SAV2 class is not calling the bundler, the BaseSmartAccount class calls the bundler and it should have the context to that file.

livingrockrises (approved):
approving for publishing in devx

arcticfloyd1984 (approved):

livingrockrises (approved):

PR #305, opened 2023-09-27
feat paymaster call for buildUserOp

AmanRaj1608:
Now the buildUSerOp will call the paymaster by default and if the skipEstimation is off it will call bundler as usual.

The code to buildUSerOp with paymasterEstimationData
```ts
  const userOp = await biconomySmartAccount.buildUserOp([transaction], {
    // skipBundlerGasEstimation: true,
    paymasterServiceData: {
      mode: PaymasterMode.SPONSORED
    },
  });
  console.log(`userOp: ${JSON.stringify(userOp, null, "\t")}`);
```


packages/paymaster/src/utils/Types.ts
```diff
@@ -112,3 +112,20 @@ export enum PaymasterMode {
...
+
+// Converted to JsonRpcResponse with strict type
+export type EstimateUserOpGasResponse = {
+  jsonrpc: string;
+  id: number;
+  result: UserOpGasResponse;
+  error?: JsonRpcError;
+};
+
+export type UserOpGasResponse = {
```
livingrockrises, line 124: why don't we just edit above?
also if there will be new api version we wouldnt know which one is being called and what to expect in response

packages/paymaster/src/utils/HelperFunction.ts
```diff
@@ -0,0 +1,24 @@
+import { UserOperation } from "@biconomy/core-types";
+import { BigNumber } from "ethers";
+
+export const transformUserOP = (userOp: UserOperation): UserOperation => {
```
livingrockrises, line 4: we didn't need this before? 
at which point we will need to transform these now. converting all to hex string

packages/paymaster/src/utils/HelperFunction.ts
```diff
@@ -0,0 +1,24 @@
...
+      "nonce",
+      "callGasLimit",
+      "verificationGasLimit",
+      "preVerificationGas",
+      "maxFeePerGas",
+      "maxPriorityFeePerGas",
+    ];
+    for (const key of keys) {
+      if (userOperation[key] && userOperation[key] !== "0") {
+        userOperation[key] = BigNumber.from(userOp[key]).toHexString();
```
livingrockrises, line 17: if it was 0 then it would remain a number. what is 0 check doing here?

packages/paymaster/src/BiconomyPaymaster.ts
```diff
@@ -35,6 +38,34 @@ export class BiconomyPaymaster implements IHybridPaymaster<SponsorUserOperationD
...
+        id: getTimestampInSeconds(),
+        jsonrpc: "2.0",
+      },
+    });
+
+    const userOpGasResponse = response.result;
+    for (const key in userOpGasResponse) {
+      if (key === "maxFeePerGas" || key === "maxPriorityFeePerGas") continue;
+      if (!userOpGasResponse[key as keyof UserOpGasResponse]) {
+        throw new Error(`Got undefined ${key} from bundler`);
```
livingrockrises, line 63: not from bundler as we are calling a paymaster here

packages/account/src/BaseSmartAccount.ts
```diff
@@ -242,8 +242,8 @@ export abstract class BaseSmartAccount implements IBaseSmartAccount {
       delete userOp.maxFeePerGas;
       delete userOp.maxPriorityFeePerGas;
       // Making call to bundler to get gas estimations for userOp
-      const { callGasLimit, verificationGasLimit, preVerificationGas, maxFeePerGas, maxPriorityFeePerGas } =
-        await this.bundler.estimateUserOpGas(userOp);
+      const { callGasLimit, verificationGasLimit, preVerificationGas, maxFeePerGas, maxPriorityFeePerGas, paymasterAndData } =
+        await this.paymaster.estimateUserOpGas(userOp);
```
livingrockrises, line 246: won't this be done as part of pm_sponsorUserOperation later anyways?
I thought the point was removing any estimations (default behavior) in this function call

packages/paymaster/src/interfaces/IPaymaster.ts
```diff
@@ -1,8 +1,9 @@
 import { UserOperation } from "@biconomy/core-types";
-import { PaymasterAndDataResponse } from "../utils/Types";
+import { PaymasterAndDataResponse, UserOpGasResponse } from "../utils/Types";
 
 export interface IPaymaster {
   // Implementing class may add extra parameter (for example paymasterServiceData with it's own type) in below function signature
+  estimateUserOpGas(_userOp: Partial<UserOperation>): Promise<UserOpGasResponse>;
```
livingrockrises, line 6: we shouldn't add this anyway in IPaymaster. Could be in IHybridPaymaster after we identify need

packages/paymaster/src/BiconomyPaymaster.ts
```diff
@@ -35,6 +38,34 @@ export class BiconomyPaymaster implements IHybridPaymaster<SponsorUserOperationD
...
+   */
+  async estimateUserOpGas(userOp: UserOperation): Promise<UserOpGasResponse> {
+    userOp = transformUserOP(userOp);
+    Logger.log("userOp sending for fee estimate ", userOp);
+
+    const response: EstimateUserOpGasResponse = await sendRequest({
+      url: `${this.paymasterConfig.paymasterUrl}`,
+      method: HttpMethod.Post,
+      body: {
+        method: "pm_sponsorship",
```
livingrockrises, line 52: @arcticfloyd1984 I mean here

livingrockrises (commented):
left comments

packages/paymaster/src/utils/HelperFunction.ts
AmanRaj1608: I saw we do this exact thing in bundler so kept the same here in bundler also as the types are same

packages/paymaster/src/interfaces/IPaymaster.ts
AmanRaj1608: Okay removed this, but now when I am calling the this.paymaster I have to set the type to HybridPaymaster in BaseAccount... Which also I feel not correct as sdk will not be generic. Let me know what you think

packages/paymaster/src/utils/Types.ts
AmanRaj1608: updated

livingrockrises (approved):

packages/account/src/BaseSmartAccount.ts
```diff
@@ -222,8 +236,10 @@ export abstract class BaseSmartAccount implements IBaseSmartAccount {
     return userOp;
   }
 
+  // TODO // Should make this a Dto
   async estimateUserOpGas(
     userOp: Partial<UserOperation>,
+    gasless: boolean,
     overrides?: Overrides,
     skipBundlerGasEstimation?: boolean,
     paymasterServiceData?: SponsorUserOperationDto,
```
AmanRaj1608: I dont think we should change this, as it will breaking change

Can do it like first and second param same as before third is boolean skipBundlerGasEstimation which can be renamed to gasless as both are doing same thing.

AmanRaj1608: nvm just checked later commits

arcticfloyd1984 (approved):

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -256,25 +256,57 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
...
+      }
+      return nonce;
+    })();
+
+    const gasFeeValues = {
+      maxFeePerGas: overrides?.maxFeePerGas,
+      maxPriorityFeePerGas: overrides?.maxPriorityFeePerGas,
+    };
+    // Estimate max fee per gas and max priority fee per gas
+    const getGasFeeValues = (async () => {
```
tomarsachin2271, line 287: make these functions (nonceFetchPromise and gasFeeValues) outside of buildUserOp as inline functions can cause performance issues as every time the outerfunction is called, the new instance of inner function is created in memory.

Rename functions to getNonce and getGasFeeValues

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -321,17 +322,51 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
...
+        }
+      } catch (error) {
+        // Not throwing this error as nonce would be 0 if this.getNonce() throw exception, which is expected flow for undeployed account
+      }
       return nonce;
     })();
     const initCodeFetchPromise = this.getInitCode();
     const dummySignatureFetchPromise = this.getDummySignature(buildUseropDto?.params);
 
+    const gasFeeValues = {
```
tomarsachin2271, line 341: Remove thes inner functions to outside of this buildUserOp method and rename methods to

getGasFeeValues
getNonce

packages/paymaster/src/BiconomyPaymaster.ts
```diff
@@ -327,19 +331,24 @@ export class BiconomyPaymaster implements IHybridPaymaster<SponsorUserOperationD
         const preVerificationGas = response.result.preVerificationGas;
         const verificationGasLimit = response.result.verificationGasLimit;
         const callGasLimit = response.result.callGasLimit;
+        const maxPriorityFeePerGas = response.result.maxPriorityFeePerGas;
```
tomarsachin2271, line 334: I don't think now paymaster service returns the maxFeePerGas or maxPriorityFeePerGas data.

packages/account/src/BiconomySmartAccountV2.ts
AmanRaj1608: we already have a `getNonce` in SmartAccountV2, this method was added just to call that method async in anonmyouse function so the `getGasFeeValues` and `getNonce` are fetched parallely.

AmanRaj1608: Should I rename it to something else then?

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -256,6 +256,37 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
     return "0x";
   }
 
+  private async getNonce(): Promise<BigNumber> {
+    let nonce = BigNumber.from(0);
+    try {
+      nonce = await this.nonce();
+    } catch (error) {
+      // Not throwing this error as nonce would be 0 if this.nonce() throw exception, which is expected flow for undeployed account
```
tomarsachin2271, line 264: Add a logger here.
"Error while getting nonce for the account. This is expected for undeployed accounts"

packages/account/src/BiconomySmartAccount.ts
```diff
@@ -256,6 +256,37 @@ export class BiconomySmartAccount extends SmartAccount implements IBiconomySmart
...
+    };
+    try {
+      if (this.bundler && !gasFeeValues.maxFeePerGas && !gasFeeValues.maxPriorityFeePerGas && (skipBundlerGasEstimation ?? true)) {
+        const gasFeeEstimation = await this.bundler.getGasFeeValues();
+        gasFeeValues.maxFeePerGas = gasFeeEstimation.maxFeePerGas;
+        gasFeeValues.maxPriorityFeePerGas = gasFeeEstimation.maxPriorityFeePerGas;
+      }
+      return gasFeeValues;
+    } catch (error: any) {
+      Logger.error("Provided bundler might not have this endpoint", error);
```
tomarsachin2271, line 285: Better logging message should be used.

"Error while getting gasFeeValues from bundler. Provided bundler might not have getGasFeeValues endpoint"

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -315,53 +317,55 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
...
+    let nonce = BigNumber.from(0);
+    try {
+      if (nonceOptions?.nonceOverride) {
+        nonce = BigNumber.from(nonceOptions?.nonceOverride);
+      } else {
+        const _nonceSpace = nonceOptions?.nonceKey ?? 0;
+        nonce = await this.getNonce(_nonceSpace);
+      }
+    } catch (error) {
+      // Not throwing this error as nonce would be 0 if this.getNonce() throw exception, which is expected flow for undeployed account
```
tomarsachin2271, line 330: Add same logger as mentioned in earlier comments.

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -315,53 +317,55 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
...
+    };
+    try {
+      if (this.bundler && !gasFeeValues.maxFeePerGas && !gasFeeValues.maxPriorityFeePerGas && (skipBundlerGasEstimation ?? true)) {
+        const gasFeeEstimation = await this.bundler.getGasFeeValues();
+        gasFeeValues.maxFeePerGas = gasFeeEstimation.maxFeePerGas;
+        gasFeeValues.maxPriorityFeePerGas = gasFeeEstimation.maxPriorityFeePerGas;
+      }
+      return gasFeeValues;
+    } catch (error: any) {
+      Logger.error("Provided bundler might not have this endpoint", error);
```
tomarsachin2271, line 351: As mentioned in earlier comments, use better logging.

tomarsachin2271 (approved):

PR #315, opened 2023-10-11
setup: devs can optionally just rely on biconomy_getUserOpStatus

livingrockrises:
# Summary

Related Issue: # (issue number)

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

- [ ] My code follows this project's style guidelines
- [ ] I've reviewed my own code

# Additional Information

# Branch Naming


AmanRaj1608 (approved):
LGTM

packages/bundler/src/Bundler.ts
```diff
@@ -154,6 +238,28 @@ export class Bundler implements IBundler {
     return userOpReceipt;
   }
 
+  /**
+   *
+   * @param userOpHash
+   * @description This function will return userOpReceipt for a given userOpHash
+   * @returns Promise<UserOpReceipt>
+   */
+  async getUserOpStatus(userOpHash: string): Promise<UserOpStatus> {
```
tomarsachin2271, line 247: Add this method to the IBundler interface as well.

packages/bundler/src/Bundler.ts
```diff
@@ -105,27 +133,83 @@ export class Bundler implements IBundler {
...
         return new Promise<UserOpReceipt>((resolve, reject) => {
-          const intervalId = setInterval(async () => {
-            try {
-              const userOpResponse = await this.getUserOpReceipt(sendUserOperationResponse.result);
-              if (userOpResponse && userOpResponse.receipt && userOpResponse.receipt.blockNumber) {
-                if (confirmations) {
-                  const latestBlock = await provider.getBlockNumber();
-                  const confirmedBlocks = latestBlock - userOpResponse.receipt.blockNumber;
-                  if (confirmations >= confirmedBlocks) {
+          if (this.UserOpReceiptIntervals && chainId in this.UserOpReceiptIntervals) {
```
tomarsachin2271, line 141: Just like we defined a default value of 30seconds for maxDuration, we can also define the default value for UserOpReceiptIntervals as well and remove this If condition.. to not have hard dependency on sdk to support new network.

packages/bundler/src/Bundler.ts
```diff
@@ -105,27 +133,83 @@ export class Bundler implements IBundler {
       userOpHash: sendUserOperationResponse.result,
       wait: (confirmations?: number): Promise<UserOpReceipt> => {
```
tomarsachin2271, line 134: Give a default value of confirmations = 0 in the method

packages/bundler/src/Bundler.ts
```diff
@@ -105,27 +133,83 @@ export class Bundler implements IBundler {
...
                   }
+                } catch (error) {
+                  clearInterval(intervalId);
+                  reject(error);
+                }
+
+                totalDuration += intervalValue;
+                if (totalDuration >= maxDuration) {
+                  clearInterval(intervalId);
+                  reject(new Error("Exceeded maximum duration"));
```
tomarsachin2271, line 167: Change the message to 
`Exceeded maximum duration (${maxDuration/1000} sec) waiting to get receipt for userOpHash ${sendUserOperationResponse.result}. Try getting the receipt manually using eth_getUserOperationReceipt rpc method on bundler`

packages/bundler/src/Bundler.ts
AmanRaj1608: the logic is already like this, so if confirmations is not defined or 0 it will just resolve and send back receipt
```ts
if (confirmations) {
  const latestBlock = await provider.getBlockNumber();
  const confirmedBlocks = latestBlock - userOpResponse.receipt.blockNumber;
  if (confirmations >= confirmedBlocks) {
    clearInterval(intervalId);
    resolve(userOpResponse);
  }
}

clearInterval(intervalId);
resolve(userOpResponse);
```

packages/account/src/BaseSmartAccount.ts
```diff
@@ -270,6 +268,9 @@ export abstract class BaseSmartAccount implements IBaseSmartAccount {
         finalUserOp.callGasLimit = callGasLimit ?? userOp.callGasLimit;
         finalUserOp.preVerificationGas = preVerificationGas ?? userOp.preVerificationGas;
         finalUserOp.paymasterAndData = paymasterAndData ?? userOp.paymasterAndData;
+        if (paymasterServiceData?.mode === PaymasterMode.ERC20) {
```
tomarsachin2271, line 271: Didn't understand this check? If mode is ERC20 why we setting paymasterAndData to be 0x?

AmanRaj1608: Because the ideal flow for token paymaster is like this. The token paymaster expects fee token which we get later and then again call the `buildTokenPaymasterUserOp`

- call bundler for max value
- call paymaster for gas value and paymasterAndData (of sponsored)
	- set paymasterAndData to 0x (as it is needed to be override later)
- call paymaster for fee quote with partialUserOp
- call paymaster for gas values and paymasterAndData (of token)
- call bundler to send tx

tomarsachin2271 (approved):
Looks good

AmanRaj1608:
Reverted the last commit as per discussion with @livingrockrises
The concern is user might not have the funding id in case of token paymaster and it will result in error.

PR #316, opened 2023-10-13
SMA 172 Account v2 aa-sdk changes

AmanRaj1608:
### Major changes

- using viem chainId
- using isHex for null checks
- using js bigint and tyoe BigNumberish from aa-sdk not ethers (major change which can lead to some edge cases)
- getDummySignature cant be used so using getDummySignatures
- using viem public client instead of ethers provider
- using encodeFunctionData instead of ethers populateMethods


0xTomDaniel:
Are we getting viem support? Most other AA SDKs chose to support viem by default.

AmanRaj1608:
> Are we getting viem support? Most other AA SDKs chose to support viem by default.

@0xTomDaniel Yes we are revamping the codebase to support viem, making it more lightweight and faster.

AmanRaj1608:
NOTE: 
- We require node version >18 as viem uses fetch inside.
- Removed the tests as now there is no way to build userOp offline. Need a way to setup bundler tests.

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -78,7 +78,42 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
   activeValidationModule!: BaseValidationModule;
 
   private constructor(readonly biconomySmartAccountConfig: BiconomySmartAccountV2Config) {
-    super(biconomySmartAccountConfig);
+    super({
+      ...biconomySmartAccountConfig,
+      chain: getChain(biconomySmartAccountConfig.chainId),
+      rpcClient: biconomySmartAccountConfig.rpcUrl || (RPC_PROVIDER_URLS[biconomySmartAccountConfig.chainId] as string),
+      entryPointAddress: biconomySmartAccountConfig.entryPointAddress as Hex,
```
VGabriel45, line 85: Here we can also use the default entry point address if not provided in biconomySmartAccountConfig like we do for factoryAddress

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -78,7 +78,42 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
...
+    this.defaultFallbackHandlerAddress = defaultFallbackHandlerAddress;
+
+    this.provider = createPublicClient({
+      chain: getChain(biconomySmartAccountConfig.chainId),
+      transport: http(biconomySmartAccountConfig.rpcUrl || (RPC_PROVIDER_URLS[biconomySmartAccountConfig.chainId] as string)),
+    });
+
+    this.scanForUpgradedAccountsFromV1 = biconomySmartAccountConfig.scanForUpgradedAccountsFromV1 ?? false;
+    this.maxIndexForScan = biconomySmartAccountConfig.maxIndexForScan ?? 10;
+    // REVIEW: removed the node client
```
VGabriel45, line 116: Why was node client removed ?

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -614,125 +711,98 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
...
+      to: await this.getAddress(),
+      value: "0x00",
+      data: callData,
     };
     return tx;
   }
 
-  async disableModule(prevModule: string, moduleAddress: string): Promise<UserOpResponse> {
-    const tx: Transaction = await this.getDisableModuleData(prevModule, moduleAddress);
+  async disableModule(preModule: Hex, moduleAddress: Hex): Promise<UserOpResponse> {
```
VGabriel45, line 776: Small typo here ? "preModule" should be "prevModule" ?

packages/account/src/abi/AccountResolver.ts
```diff
@@ -0,0 +1,106 @@
+export const AccountResolverAbi = [
```
livingrockrises, line 1: rename this to AddressResolver and in file name also

packages/paymaster/src/utils/Types.ts
```diff
@@ -103,11 +104,9 @@ export type FeeQuotesOrDataResponse = {
 
 export type PaymasterAndDataResponse = {
   paymasterAndData: string;
-  preVerificationGas?: BigNumberish;
-  verificationGasLimit?: BigNumberish;
-  callGasLimit?: BigNumberish;
-  maxPriorityFeePerGas?: BigNumberish;
-  maxFeePerGas?: BigNumberish;
+  preVerificationGas?: string;
```
livingrockrises, line 107: do we receive strings from our paymaster service?

livingrockrises:
can you also link of example code here

packages/account/src/utils/Types.ts
```diff
@@ -1,11 +1,11 @@
 import { Signer } from "ethers";
```
VGabriel45, line 1: Shouldn't we use viem now ? So this should be compatible with viem client now not ethers v5

packages/account/src/BiconomySmartAccountV2.ts
AmanRaj1608: It was required for v1 client where we used to call a backend service. We don't need any backend service as of now so removing it.

joepegler (approved):
LGTM

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -94,127 +129,55 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
...
-    }
-
-    const defaultFallbackHandlerAddress =
-      instance.factoryAddress === DEFAULT_BICONOMY_FACTORY_ADDRESS
-        ? DEFAULT_FALLBACK_HANDLER_ADDRESS
-        : biconomySmartAccountConfig.defaultFallbackHandler;
-    if (!defaultFallbackHandlerAddress) {
-      throw new Error("Default Fallback Handler address is not provided");
-    }
-    instance.accountAddress = biconomySmartAccountConfig.senderAddress ?? undefined;
```
livingrockrises, line 112: we still need these in init right?

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -94,127 +129,55 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
...
 
-    const { rpcUrl, nodeClientUrl } = biconomySmartAccountConfig;
-
-    if (rpcUrl) {
-      instance.provider = new JsonRpcProvider(rpcUrl);
-    }
-
-    instance.nodeClient = new NodeClient({ txServiceUrl: nodeClientUrl ?? NODE_CLIENT_URL });
-
-    instance.scanForUpgradedAccountsFromV1 = biconomySmartAccountConfig.scanForUpgradedAccountsFromV1 ?? false;
```
livingrockrises, line 137: this part too if not done in init where it is set?

packages/paymaster/src/utils/Types.ts
AmanRaj1608: updated

packages/account/src/utils/Types.ts
AmanRaj1608: resolved

packages/account/src/BiconomySmartAccountV2.ts
AmanRaj1608: Instead of doing these in the init/create have moved it to private constructor.
Keeping the `create` method to do async things. So I call constructor where I do all this and then call async tasks like this

```ts
if (biconomySmartAccountConfig.defaultValidationModule) {
      instance.defaultValidationModule = biconomySmartAccountConfig.defaultValidationModule;
    } else {
      instance.defaultValidationModule = await ECDSAOwnershipValidationModule.create({
    });
}
```

packages/account/package.json
```diff
@@ -34,22 +34,16 @@
     "access": "public"
   },
   "devDependencies": {
-    "nock": "^13.2.9"
+    "nock": "^13.2.9",
+    "viem": "^1.19.11"
```
Aboudjem, line 38: keep it (viem) either on dep or devdep

livingrockrises (approved):

PR #332, opened 2023-11-22
Update import paths

AmanRaj1608:


packages/modules/src/session-storage/SessionLocalStorage.ts
```diff
@@ -1,5 +1,5 @@
 import { Wallet, Signer } from "ethers";
-import { ISessionStorage, SessionLeafNode, SessionSearchParam, SessionStatus } from "interfaces/ISessionStorage";
+import { ISessionStorage, SessionLeafNode, SessionSearchParam, SessionStatus } from "../interfaces/ISessionStorage";
```
basememara, line 2: Curious if this was causing an issue or if this was for consistency? Because it doesn't look like the relative path is needed since `baseUrl` is set to `src` in `tsconfig.json`. Will this require changing if the folder structure is changed in the future?

Maybe creating a special base path prefix like Next.js does would make this clearer, but I don't think that's necessary either if we always assume `src` is the root path.

```ts
// tsconfig.json
{
  "compilerOptions": {
    ...
    "paths": {
      "@/*": ["./*"]
    }
  }
}

// SessionLocalStorage.ts
import { ISessionStorage,... } from "@/interfaces/ISessionStorage";
```

livingrockrises (approved):
LGTM

arcticfloyd1984 (approved):

packages/modules/src/session-storage/SessionLocalStorage.ts
AmanRaj1608: This is for consistency only. We do have  configured the baseUrl in tsConfig. But someone was still getting error on nodejs node version 20. So making this consistent on overall project.

basememara: Interesting I wonder what's in v20 that's causing this and will keep a look out, thanks for the insight!

PR #334, opened 2023-12-04
Feat/devx 336 abstract paymaster

VGabriel45:
# Summary

Related Issue: # (DEVX-336)

## Change Type

- [ ] Bug Fix
- [x] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

# Additional Information

The context of this PR is to optimize developer experience of the SDK.
This PR will remove the need for the user/developer to create the Paymaster instance and then provide it to the "create()" method config. Paymaster can be easily setup by only providing an "apiKey" in the config.

Old method of creating instance and providing it to the config by the user/developer is still available for compatibility reasons.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,11 +52,11 @@
     ]
   },
   "dependencies": {
+    "dotenv": "^16.3.1",
     "node-gyp": "^9.4.0",
     "typescript": "^5.2.2"
   },
   "devDependencies": {
-    "ganache": "^7.9.1",
     "@types/debug": "^4.1.9",
     "@types/jest": "^29.5.4",
     "@typescript-eslint/eslint-plugin": "^6.7.0",
@@ -68,6 +68,7 @@
     "eslint-plugin-import": "^2.28.1",
     "eslint-plugin-prettier": "^5.0.0",
     "eslint-plugin-security": "^1.7.1",
+    "ganache": "^7.9.1",
     "hardhat": "^2.17.3",
     "jest": "^29.7.0",
     "lerna": "^7.2.0",
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     ]
   },
   "dependencies": {
-    "dotenv": "^16.3.1",
     "node-gyp": "^9.4.0",
     "typescript": "^5.2.2"
   },
--- a/packages/account/src/BaseSmartAccount.ts
+++ b/packages/account/src/BaseSmartAccount.ts
@@ -52,12 +52,6 @@ export abstract class BaseSmartAccount implements IBaseSmartAccount {
 
     this.provider = _smartAccountConfig.provider ?? new JsonRpcProvider(RPC_PROVIDER_URLS[this.chainId]);
 
-    if (_smartAccountConfig.apiKey) {
-      this.paymaster = new BiconomyPaymaster({
-        paymasterUrl: `https://paymaster.biconomy.io/api/v1/${_smartAccountConfig.chainId}/${_smartAccountConfig.apiKey}`,
-      });
-    }
-
     // Create an instance of the EntryPoint contract using the provided address and provider (facory "connect" contract address)
     // Then, set the transaction's sender ("from" address) to the zero address (AddressZero). (contract "connect" from address)
     this.entryPoint = EntryPoint__factory.connect(this.entryPointAddress, this.provider).connect(ethers.constants.AddressZero);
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -52,11 +52,11 @@
     ]
   },
   "dependencies": {
+    "dotenv": "^16.3.1",
```
livingrockrises, line 55: why do we need dotenv here?

packages/account/src/BaseSmartAccount.ts
```diff
@@ -47,7 +47,6 @@ export abstract class BaseSmartAccount implements IBaseSmartAccount {
     this.overheads = _smartAccountConfig.overheads;
     this.entryPointAddress = _smartAccountConfig.entryPointAddress ?? DEFAULT_ENTRYPOINT_ADDRESS;
     this.accountAddress = _smartAccountConfig.accountAddress;
-    this.paymaster = _smartAccountConfig.paymaster;
```
livingrockrises, line 50: this should be assigned if provided right

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -84,6 +84,12 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
     const instance = new BiconomySmartAccountV2(biconomySmartAccountConfig);
     instance.factoryAddress = biconomySmartAccountConfig.factoryAddress ?? DEFAULT_BICONOMY_FACTORY_ADDRESS; // This would be fetched from V2
 
+    if (biconomySmartAccountConfig.apiKey) {
```
livingrockrises, line 87: if biconomySmartAccountConfig has paymaster instance then assign it 
if it has apiKey then do this 

also let's rename apiKey to something like biconomyPaymasterApiKey

packages/account/src/utils/Types.ts
```diff
@@ -69,6 +69,7 @@ export interface BiconomySmartAccountV2Config extends BaseSmartAccountConfig {
   senderAddress?: string;
   implementationAddress?: string;
   defaultFallbackHandler?: string;
+  apiKey?: string;
```
livingrockrises, line 72: rename for it to be more readable

packages/account/tests/SmartAccountV2-Abstract-Paymaster.local.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    ecdsaModule = await new ECDSAOwnershipRegistryModule_v100__factory(signer).deploy();
+
+    module1 = await ECDSAOwnershipValidationModule.create({
+      signer: owner,
+      moduleAddress: ecdsaModule.address,
+    });
+
+    console.log("provider url ", provider.connection.url);
+
+    recipient = await new SampleRecipient__factory(signer).deploy();
```
livingrockrises, line 58: you don't need this recipient here

livingrockrises (commented):
reviewed

package.json
VGabriel45: For testing, we need to provide apiKey so in the test file it is given through env.

packages/account/src/BaseSmartAccount.ts
VGabriel45: Yes, I missed that, removed it for some testing then forgot to add it back.

package.json
livingrockrises: I see. I think it should be fine if you just use hard coded key in the test file itself and then provide it to smart account config.

VGabriel45: Ok then wil remove dotenv and use hardcoded values

packages/account/src/BaseSmartAccount.ts
livingrockrises: you can resolve it once you push

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -84,6 +84,12 @@ export class BiconomySmartAccountV2 extends BaseSmartAccount {
     const instance = new BiconomySmartAccountV2(biconomySmartAccountConfig);
     instance.factoryAddress = biconomySmartAccountConfig.factoryAddress ?? DEFAULT_BICONOMY_FACTORY_ADDRESS; // This would be fetched from V2
 
+    if (biconomySmartAccountConfig.biconomyPaymasterApiKey) {
```
livingrockrises, line 87: in the else part here if config.paymaster is present then assign it to instance.paymaster

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45, line 1: @livingrockrises how often would we need to change the "v1" from the paymasterUrl here ? because right now this is hardcoded

livingrockrises: not often, if infra squad creates paymaster v2 apis then I believe other things would change too in the sdk.

livingrockrises (approved):

AmanRaj1608 (approved):

PR #337, opened 2023-12-08
Added check for entryPointAddress field

VGabriel45:
# Summary

Added conditional logic for situations when user will not provide "entryPointAddress" to the Bundler.

Related Issue: # 

## Change Type

- [x] Bug Fix
- [ ] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

### DIFF
--- a/packages/bundler/src/utils/Constants.ts
+++ b/packages/bundler/src/utils/Constants.ts
@@ -125,3 +125,5 @@ export const UserOpWaitForTxHashMaxDurationIntervals: { [key in ChainId]?: numbe
   [ChainId.CHILIZ_MAINNET]: 20000,
   [ChainId.CHILIZ_TESTNET]: 20000,
 };
+
+export const DEFAULT_ENTRYPOINT_ADDRESS = "0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789";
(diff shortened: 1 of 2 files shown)


packages/bundler/src/utils/Constants.ts
```diff
@@ -125,3 +125,5 @@ export const UserOpWaitForTxHashMaxDurationIntervals: { [key in ChainId]?: numbe
   [ChainId.CHILIZ_MAINNET]: 20000,
   [ChainId.CHILIZ_TESTNET]: 20000,
 };
+
+export const DEFAULT_ENTRYPOINT_ADDRESS = "0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789";
```
AmanRaj1608, line 129: could also import this from the baseValidationModule, but this is also fine if we don't wanna have circular dependency

AmanRaj1608 (changes requested):
Fix lint issues

packages/bundler/src/utils/Constants.ts
VGabriel45: If it's fine like this I'm gonna let it

livingrockrises (approved):
LGTM

AmanRaj1608 (approved):

PR #340, opened 2023-12-21
feat!: reduce the bundle size

AmanRaj1608:
- Removed the `common`, `core-types`, `node-client` packages.
- Remove the ethers dependency
- Removed unwanted codebase


packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -434,9 +430,8 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
       "signature",
     ];
     this.validateUserOp(userOp, requiredFields);
-    Logger.log("userOp validated");
     if (!this.bundler) throw new Error("Bundler is not provided");
-    Logger.log("userOp being sent to the bundler", userOp);
+    console.info("userOp being sent to the bundler", userOp);
```
joepegler, line 434: We're logging everything now, regardless of the environment variable right? I quite liked the logger. Did we only get rid of it because it was in the common package?

joepegler: Won't we be spamming browser consoles now?

packages/account/tests/SmartAccountV2-Abstract-Paymaster.local.spec.ts
```diff
@@ -1,102 +1,86 @@
...
-import { ChainId } from "@biconomy/core-types";
-import { ECDSAOwnershipValidationModule } from "@biconomy/modules";
-import { BaseValidationModule } from "@biconomy/modules";
-import { ECDSAOwnershipRegistryModule_v100 } from "@biconomy/common";
-import { BiconomyPaymaster } from "@biconomy/paymaster";
-import { Hex } from "viem";
-
-const provider = new ethers.providers.JsonRpcProvider("http://127.0.0.1:8545");
-const signer = provider.getSigner();
+// import { EntryPoint, EntryPoint__factory } from "@account-abstraction/contracts";
```
joepegler, line 1: Commented until we can use this right? https://github.com/bcnmy/scw-contracts/tree/main/test/bundler-integration

packages/bundler/src/Bundler.ts
```diff
@@ -139,7 +137,10 @@ export class Bundler implements IBundler {
     const response: UserOpResponse = {
       userOpHash: sendUserOperationResponse.result,
       wait: (confirmations?: number): Promise<UserOpReceipt> => {
-        const provider = new JsonRpcProvider(RPC_PROVIDER_URLS[chainId]);
+        const providerClient = createPublicClient({
+          chain: getChain(chainId) || mainnet,
```
joepegler, line 141: Should this default to mainnet here if there's no chainId? Might be better to just error?

AmanRaj1608: Makes sense, updating

packages/account/src/BiconomySmartAccountV2.ts
AmanRaj1608: Currently moved to console for time being, as I wanted to remove the dependency of common package.

Need to raise another PR where we can either use the same Logger in utils or use another logger service. Keeped the console so we get what to repalce

VGabriel45 (approved):

PR #341, opened 2023-12-22
Feat/devx 362 abstract required bundler

VGabriel45:
# Summary

Abstracted the need for the client to create an instance of Bundler and assign it to the smart account.
Added a light way of assigning a Bundler by just providing a "bundlerUrl" param.

## Change Type

- [ ] Bug Fix
- [x] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist


package.json
```diff
@@ -52,6 +52,7 @@
     ]
   },
   "dependencies": {
+    "lru-cache": "^10.1.0",
```
AmanRaj1608, line 55: This is only being used in account package and have been already added there any reason adding in core package?

AmanRaj1608 (approved):
Need to fix merge conflicts, other than that LGTM

livingrockrises:
can you please resolve the conflicts

packages/account/src/BaseSmartAccount.ts
```diff
@@ -47,9 +47,18 @@ export abstract class BaseSmartAccount implements IBaseSmartAccount {
     this.overheads = _smartAccountConfig.overheads;
     this.entryPointAddress = _smartAccountConfig.entryPointAddress ?? DEFAULT_ENTRYPOINT_ADDRESS;
     this.accountAddress = _smartAccountConfig.accountAddress;
-    this.bundler = _smartAccountConfig.bundler;
+
     this.chainId = _smartAccountConfig.chainId;
 
+    if (_smartAccountConfig.bundlerUrl) {
```
livingrockrises, line 53: we can also do conditional props for these? or an overkill
@AmanRaj1608 @VGabriel45 @joepegler

packages/account/src/utils/Types.ts
```diff
@@ -89,6 +88,14 @@ type ConditionalValidationProps = RequireAtLeastOne<
   "defaultValidationModule" | "signer"
 >;
 
+type ConditionalBundlerProps = RequireAtLeastOne<
```
livingrockrises, line 91: I see you already have conditional props. would there be a case when bundler is not used at all?
if someone is using only paymaster service for estimations and just building userops. 
@AmanRaj1608

livingrockrises (commented):
resolve conflicts and check comments

packages/account/src/utils/Types.ts
VGabriel45: As we discussed yesterday, I believe bundler should be required as you cannot do much without it and it will also help us to abstract things like chain id in the future. Clients that currently create smart accounts without bundler will need to update the codebase.

livingrockrises: we should mention this in the docs after release

livingrockrises (approved):

PR #343, opened 2023-12-27
Feat/devx 405 abstract chainId

VGabriel45:
# Summary

The scope of this PR is to abstract away the need to provide a chain id manually, we can extract the chain id from bundler url as bundler is now required.

Related Issue: #DEVX-405

## Change Type

- [ ] Bug Fix
- [x] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

### DIFF
--- a/packages/account/src/BaseSmartAccount.ts
+++ b/packages/account/src/BaseSmartAccount.ts
@@ -48,12 +48,12 @@ export abstract class BaseSmartAccount implements IBaseSmartAccount {
     this.entryPointAddress = _smartAccountConfig.entryPointAddress ?? DEFAULT_ENTRYPOINT_ADDRESS;
     this.accountAddress = _smartAccountConfig.accountAddress;
 
-    this.chainId = _smartAccountConfig.chainId;
+    this.chainId = _smartAccountConfig.chainId!;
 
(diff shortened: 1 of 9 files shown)


VGabriel45:
The PR also contains latest "develop" branch updates

joepegler:
Hmmm, I like this but I think I'd prefer to set the bundlerUrl (to our own bundler) only if the chainId is missing. So the chainId becomes mandatory if the bunderlUrl is missing, as the only dynamic part of our bundlerUrl is the chainId. 

I think removing the chainId from the config entirely could cause confusion as it's a param I'd expect to see. 

TLDR: Inferring the bunderUrl from the chainId would make more sense than inferring the chainId from the bundlerUrl.

packages/account/src/BaseSmartAccount.ts
AmanRaj1608, line 1: We don't need this in v4

AmanRaj1608 (changes requested):
Do take a back merge from v4

livingrockrises:
resolve conflicts

PR #351, opened 2024-01-04
Feat(DEVX-410): E2E Tests

joepegler:
# Summary

[Notion Doc](https://www.notion.so/biconomy/Testing-9e379ff0c32147e3bdf62b4c31d62759)
Related Issue: [#DEVX-410](https://linear.app/biconomy/issue/DEVX-410)

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [x] Other

# Checklist

# Additional Information

I've added some additional aliases for devEx purposes, and then used them in the tests

createSmartWalletAccount -> BiconomySmartAccountV2.create
smartWallet.sendTransaction(s) -> BiconomySmartAccountV2.buildUserOp + BiconomySmartAccountV2.sendUserOp

I've also imported and then re-exported the Bundler and Paymaster packages from the accounts package, so that mostly everything can be important from "@biconomy/account"


packages/account/src/aliases.ts
```diff
@@ -0,0 +1,7 @@
+import { BiconomySmartAccountV2 } from "./BiconomySmartAccountV2";
+import { BiconomySmartAccountV2Config } from "./utils/Types";
+
+export { BiconomyPaymaster as Paymaster, IPaymaster } from "@biconomy/paymaster";
+export { Bundler, IBundler } from "@biconomy/bundler";
+export const createSmartWalletClient = BiconomySmartAccountV2.create;
+export type SmartWalletConfig = BiconomySmartAccountV2Config;
```
AmanRaj1608, line 7: > export const createSmartWalletClient = BiconomySmartAccountV2.create;

this would be very helpful.

Also would be better to do this in index.ts rather creating a new file, just to have a minimal codebase.

packages/account/tests/account.e2e.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+    };
+
+    const epHash = await publicClient.readContract({
+      address: entryPointAddress as Hex,
+      abi: [
+        {
+          inputs: [
+            {
+              components: [
+                { internalType: "address", name: "sender", type: "address" },
```
AmanRaj1608, line 173: Can reduce loc here with parseAbi from viem, but need to confirm as the params are not straight forward.
Just check if it can be otherwise looks good

```ts
const parsedAbi = parseAbi(["function getUserOpHash(...)"]);
const epContract = getContract({
   address: entryPointAddress as Hex,
   abi: parsedAbi,
   publicClient,
});
const count = await epContract.read.getUserOpHash([userOp]);
```

AmanRaj1608 (commented):
Overall I have these few concerns
- What if there are multiple PR's opening at same time, test might fail if the nonce is incorrect.
- Hard to manage the paymaster with balance.
- For non-gasless tests would need to make sure and have the funds in EOA.

Latency is also one of my concern as build time increases it will cost for on CI/CD infra. In this PR I can see it is 1m. But if we add more tests this can go up really fast.

AmanRaj1608 (approved):
LGTM!

AmanRaj1608:
Even tho I approved this it's still not making the build. Looks like it needs all approvals.
Do we have any other `github.event.review.state` ? @joepegler

AmanRaj1608 (approved):

PR #352, opened 2024-01-04
fix: bundler abstraction in v4

AmanRaj1608:
- Updating the bundler abstraction for v4
- Fixed the token paymaster bigint error throw
- gitignore update

### DIFF
--- a/packages/account/src/BiconomySmartAccountV2.ts
+++ b/packages/account/src/BiconomySmartAccountV2.ts
@@ -22,7 +22,7 @@ import { BaseSmartContractAccount, getChain, type BigNumberish, type UserOperati
 import { isNullOrUndefined, packUserOp } from "./utils/Utils";
 import { BaseValidationModule, ModuleInfo, SendUserOpParams, ECDSAOwnershipValidationModule } from "@biconomy/modules";
 import { IHybridPaymaster, IPaymaster, BiconomyPaymaster, SponsorUserOperationDto } from "@biconomy/paymaster";
-import { IBundler, UserOpResponse } from "@biconomy/bundler";
+import { Bundler, IBundler, UserOpResponse } from "@biconomy/bundler";
 import {
   BiconomyTokenPaymasterRequest,
   BiconomySmartAccountV2Config,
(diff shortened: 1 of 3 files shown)


packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -99,6 +99,15 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
       this.paymaster = biconomySmartAccountConfig.paymaster;
     }
 
+    if (biconomySmartAccountConfig.bundlerUrl) {
```
joepegler, line 102: I know it's not in this PR, but is there a strong reason why we're not doing the ecdsa validation module abstraction on the instantiated Class (in the static `create` call) instead of in the constructor like you've done here for the bundler?  

https://github.com/bcnmy/biconomy-client-sdk/blob/591bbb4e37774b16cbe801d583d31b3a14608bc1/packages/account/src/BiconomySmartAccountV2.ts#L142

joepegler (commented):
LGTM 👍

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45: I think bundler creation has higher priority than the validation module so it should be done in the constructor, we can also add the validation module there but might be better to keep the constructor as small as possible.

AmanRaj1608: I try to do everything which is synchronously in constructor. As we can't do any async operation in constructor we have this `.create` method where we can do async stuff and client doesn't need to first generate constructor and then call async method.

So client calls .create which will initialise constructor with all things, only async operation will be done in .create

joepegler: That makes sense, but you can do the async stuff and then feed the result into the constructor as the config, instead of setting it on the instance afterwards. Probably has no tangible effect, it just read a bit weird.

joepegler: Did we have to make instance.defaultValidationModule public instead of private for this pattern?

joepegler: Maybe it was public anyway, but if it were private this wouldn't work

AmanRaj1608: This was not the reason to make it private, I just made most params private which I client doesn't need.

joepegler: You could return defaultValidationModule to being private like so: 

```  
public static async create(biconomySmartAccountConfig: BiconomySmartAccountV2Config): Promise<BiconomySmartAccountV2> {
    const config = {
      ...biconomySmartAccountConfig,
      defaultValidationModule: biconomySmartAccountConfig?.defaultValidationModule,
      activeValidationModule: biconomySmartAccountConfig?.defaultValidationModule,
    };

    if (!biconomySmartAccountConfig?.defaultValidationModule) {
      const newModule = await ECDSAOwnershipValidationModule.create({
      });
      config.defaultValidationModule = newModule;
      config.activeValidationModule = newModule;
    }

    return new BiconomySmartAccountV2(config);
  }
```

PR #353, opened 2024-01-05
Feat: Add Logger

AmanRaj1608:
Instead of adding a new package for adding logger in all packages, let's reuse the Logger class.

### DIFF
--- a/packages/account/src/BiconomySmartAccountV2.ts
+++ b/packages/account/src/BiconomySmartAccountV2.ts
@@ -133,13 +133,13 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
     };
 
     // Note: If no module is provided, we will use ECDSA_OWNERSHIP as default
-    if (!biconomySmartAccountConfig?.defaultValidationModule) {
+    if (!config?.defaultValidationModule) {
       const newModule = await ECDSAOwnershipValidationModule.create({
         signer: biconomySmartAccountConfig.signer!,
       });
       config.defaultValidationModule = newModule;
-      config.activeValidationModule = newModule;
     }
+    config.activeValidationModule = config?.activeValidationModule ?? config.defaultValidationModule;
 
     return new BiconomySmartAccountV2(config);
   }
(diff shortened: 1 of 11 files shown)


packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -127,19 +128,20 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+    const config = {
+      ...biconomySmartAccountConfig,
+    };
 
     // Note: If no module is provided, we will use ECDSA_OWNERSHIP as default
-    if (biconomySmartAccountConfig.defaultValidationModule) {
-      instance.defaultValidationModule = biconomySmartAccountConfig.defaultValidationModule;
-    } else {
-      instance.defaultValidationModule = await ECDSAOwnershipValidationModule.create({
+    if (!biconomySmartAccountConfig?.defaultValidationModule) {
```
joepegler, line 136: My logic was slightly off here.

```
if (!config?.defaultValidationModule) {
  const newModule = await ECDSAOwnershipValidationModule.create({
  });
  config.defaultValidationModule = newModule;
}
config.activeValidationModule = config?.activeValidationModule ?? config.defaultValidationModule;

return new BiconomySmartAccountV2(config);
```

joepegler (changes requested):

packages/account/src/BiconomySmartAccountV2.ts
AmanRaj1608: Yeah makes sense, updated the code

joepegler (approved):

joepegler (changes requested):
You'll also need to set it in the constructor the same way you do the others: 

```    this.defaultValidationModule = biconomySmartAccountConfig.defaultValidationModule;
    this.activeValidationModule = biconomySmartAccountConfig.activeValidationModule;

```

joepegler (approved):

VGabriel45:
Can we have the Logger class just in one place and use it from there instead of adding it in every module where we need it ?

AmanRaj1608:
@VGabriel45 This was there previously in `common` package. Adding this will introduce again another package which we need to deploy. If we have it in one package and import in another it can introduce circular dependecies
account importing paymaster for paymaster class and paymaster importing account for common package.

So solution I think of
- Use an external logger class and add as dependency in all class
- Re-use code but this is much lighter than any package and does the job

VGabriel45 (approved):

PR #358, opened 2024-01-11
Ethers support

joepegler:
# Summary

Support ethers:
```
import { createSmartWalletClient } from "@biconomy/account"
import { ethers } from "ethers";

const ethersProvider = new ethers.providers.JsonRpcProvider(chain.viemChain.rpcUrls.public.http[0]);
const ethersSigner = new ethers.Wallet(privateKey, ethersProvider);

const smartWallet = await createSmartWalletClient({ signer:  ethersSigner, ... })
```
Related Issue: #DEV-425

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

- [ x My code follows this project's style guidelines
- [ x I've added tests that prove my fix is effective or my feature works

### DIFF
--- a/packages/account/package.json
+++ b/packages/account/package.json
@@ -50,6 +50,7 @@
     "@biconomy/bundler": "^3.1.2",
     "@biconomy/modules": "^3.1.2",
     "@biconomy/paymaster": "^3.1.2",
+    "ethers": "^6.9.2",
     "lru-cache": "^10.0.1",
     "viem": "^1.20.3"
   }
--- a/packages/account/src/BiconomySmartAccountV2.ts
+++ b/packages/account/src/BiconomySmartAccountV2.ts
@@ -28,9 +28,9 @@ import {
   WalletClientSigner,
 } from "@alchemy/aa-core";
 import { isNullOrUndefined, packUserOp } from "./utils/Utils";
+import { Bundler, IBundler, UserOpResponse } from "@biconomy/bundler";
 import { BaseValidationModule, ModuleInfo, SendUserOpParams, ECDSAOwnershipValidationModule } from "@biconomy/modules";
 import { IHybridPaymaster, IPaymaster, BiconomyPaymaster, SponsorUserOperationDto } from "@biconomy/paymaster";
-import { Bundler, IBundler, UserOpResponse } from "@biconomy/bundler";
 import {
   BiconomyTokenPaymasterRequest,
   BiconomySmartAccountV2Config,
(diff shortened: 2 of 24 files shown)


packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -159,18 +162,33 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
-      const isViemWalletClient = !(signer as WalletClientSigner)?.signerType;
-      if (isViemWalletClient) {
-        const walletClient = signer as WalletClient;
-        if (!walletClient.account) {
-          throw new Error("Cannot consume a viem wallet without an account");
-        }
-        if (!walletClient.chain) {
-          throw new Error("Cannot consume a viem wallet without a chainId");
+
+      const isAnAlchemySmartWallet = UNIQUE_PROPERTIES_PER_SIGNER.AlchemySmartAccountSigner in signer;
```
joepegler, line 166: Could use `instanceof SmartAccountSigner` here (and`instanceof Signer`, `instanceof WalletClient`) but not sure how consistent this check is on the instance across different package versions 🤔.

packages/account/package.json
```diff
@@ -50,6 +50,7 @@
     "@biconomy/bundler": "^3.1.2",
     "@biconomy/modules": "^3.1.2",
     "@biconomy/paymaster": "^3.1.2",
+    "ethers": "^6.9.2",
```
AmanRaj1608, line 53: Should use [@ethersproject/providers@5.7.2](https://bundlephobia.com/package/@ethersproject/providers@5.7.2) instead of [ethers](https://bundlephobia.com/package/ethers@6.10.0)

joepegler:
I've accommodate the ethersProject Signer but now I think we should add a final type to the list of signers we support: JsonRPCProvider, which calls provider.getSigner under the hood.

AmanRaj1608:
Need to take another back-merge from v4

joepegler:
> I've accommodate the ethersProject Signer but now I think we should add a final type to the list of signers we support: JsonRPCProvider, which calls provider.getSigner under the hood.

This doesn't make sense

packages/account/package.json
```diff
@@ -51,6 +51,9 @@
     "@biconomy/bundler": "^3.1.2",
     "@biconomy/modules": "^3.1.2",
     "@biconomy/paymaster": "^3.1.2",
+    "@ethersproject/abstract-signer": "^5.7.0",
+    "@ethersproject/providers": "^5.7.2",
+    "@ethersproject/wallet": "^5.7.0",
```
AmanRaj1608, line 56: We are only using the `ethersproject/providers` in tests. Can we remove this from dependency. If still needed should be there in dev dependency right?

AmanRaj1608 (changes requested):
All other changes are working, can be merged after this

PR #368, opened 2024-01-17
Abstract away chain specific data from tests

joepegler:
https://github.com/bcnmy/biconomy-client-sdk/pull/368/files#diff-d0334c69352e90b0ae3e0c802c45e1d8893550906cb6d1c0647f8840aec9fe01R16

Going forward we can add test chains to the above config file^


packages/account/tests/account.e2e.spec.ts
```diff
@@ -104,14 +106,59 @@ describe("Account Tests", () => {
     expect(newBalance).toEqual(balance);
   }, 60000);
 
+  it("Should gaslessly mint an NFT on Base base", async () => {
```
VGabriel45, line 109: No need to do this test, we should use Base testnet only in Multichain test, if NFT mint works on mumbai there is not need to test on other testnets, it will just drain more test funds from the wallets.

joepegler: removed

PR #371, opened 2024-01-18
Paymaster check and setup logic on SDK (SPONSORED | ERC20).

VGabriel45:
# Summary

This PR improves DEVX as most of the work to setup a SPONSORED or ERC20 paymaster user op is now done in the SDK.
It is not a breaking change because developers can also construct a SPONSORED or ERC20 paymaster user op the same way the did before too.

## Change Type

- [ ] Bug Fix
- [x] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist


packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -499,11 +507,75 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+          finalUserOp.callGasLimit = paymasterAndDataWithLimits.callGasLimit;
+          finalUserOp.verificationGasLimit = paymasterAndDataWithLimits.verificationGasLimit;
+          finalUserOp.preVerificationGas = paymasterAndDataWithLimits.preVerificationGas;
+          return finalUserOp;
+        } else {
+          return userOp;
+        }
+      } catch (e: any) {
+        Logger.error("Error while fetching paymaster data", e);
+        return userOp;
```
joepegler, line 566: Should we return or just throw here? Difficult to understand when / why things should fail, but are we risking users submitting wrong values by returning here when it could error instead?

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -499,11 +507,75 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+      }
+    } else {
+      Logger.error("Paymaster is not provided");
```
joepegler, line 569: Same comment as above

packages/account/src/utils/Types.ts
```diff
@@ -177,6 +177,20 @@ export type InitilizationData = {
...
+  calculateGasLimits?: boolean;
+  expiryDuration?: number;
+  webhookData?: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    [key: string]: any;
+  };
+  smartAccountInfo?: SmartAccountData;
+  feeTokenAddress?: string;
+  tokenList?: string[];
+  preferredToken?: string;
```
joepegler, line 191: Hex?

packages/account/tests/account.e2e.spec.ts
```diff
@@ -90,20 +86,73 @@ describe("Account Tests", () => {
     };
 
     const balance = (await checkBalance(publicClient, recipient, nftAddress)) as bigint;
-    const partialUserOp = await smartWallet.buildUserOp([transaction]);
 
-    const paymasterData = await paymaster.getPaymasterAndData(partialUserOp, {
-      mode: PaymasterMode.SPONSORED,
+    const maticBalanceBefore = publicClient.getBalance({ address: await smartWallet.getAddress() });
```
joepegler, line 90: There's a utils function for this in the tests directory

joepegler: checkBalance used above

packages/account/tests/account.e2e.spec.ts
```diff
@@ -90,20 +86,73 @@ describe("Account Tests", () => {
...
+
+    const response = await smartWallet.sendTransaction(transaction, { paymasterServiceData: { mode: PaymasterMode.SPONSORED } });
+
+    const userOpReceipt = await response.wait(3);
```
joepegler, line 94: Just out of curiosity, why are we waiting for 3 blocks here?

packages/account/tests/account.e2e.spec.ts
```diff
@@ -90,20 +86,73 @@ describe("Account Tests", () => {
...
+    expect(userOpReceipt.userOpHash).toBeTruthy();
+    expect(userOpReceipt.success).toBe("true");
+
+    const maticBalanceAfter = publicClient.getBalance({ address: await smartWallet.getAddress() });
```
joepegler, line 98: checkBalance helper

packages/account/tests/account.e2e.spec.ts
```diff
@@ -90,20 +86,73 @@ describe("Account Tests", () => {
...
+    const encodedCall = encodeFunctionData({
+      abi: parseAbi(["function safeMint(address _to)"]),
+      functionName: "safeMint",
+      args: [recipient],
+    });
+
+    const transaction = {
+      to: nftAddress, // NFT address
+      data: encodedCall,
+      value: 0,
```
joepegler, line 131: If you backmerge you should be able to remove the value here in a typesafe manner.

packages/account/tests/account.e2e.spec.ts
```diff
@@ -90,20 +86,73 @@ describe("Account Tests", () => {
...
+    };
+
+    const balance = (await checkBalance(publicClient, recipient, nftAddress)) as bigint;
+
+    const maticBalanceBefore = publicClient.getBalance({ address: await smartWallet.getAddress() });
```
joepegler, line 136: use helper

packages/paymaster/src/utils/Types.ts
```diff
@@ -109,10 +109,10 @@ export type PaymasterAndDataResponse = {
   callGasLimit: number;
 };
 
-export enum PaymasterMode {
-  ERC20 = "ERC20",
-  SPONSORED = "SPONSORED",
-}
+// export enum PaymasterMode {
```
joepegler, line 112: Why commented?

joepegler (dismissed):
Nice job on this. Few small tidying bits.

joepegler:
Let's make sure we touch on this before releasing, as we had a similar solution in v3 that ultimately caused the bundler to lose funds. Here are some relevant threads

- https://biconomyworkspace.slack.com/archives/C066X6E3J9Y/p1704888602987659
- https://biconomyworkspace.slack.com/archives/C03G74F8W76/p1703835030390559

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45: I think both ways are ok, I chose to return user op to not add more return type to the function.

packages/account/src/utils/Types.ts
VGabriel45: Can leave it as string for now, if I change to hex will probably need to check other places as well.

packages/account/tests/account.e2e.spec.ts
VGabriel45: No specific reason, just wanted to make sure the oser op is sent before proceeding, i can change it.

packages/paymaster/src/utils/Types.ts
VGabriel45: Good catch, forgot to uncomment these.

VGabriel45:
> Let's make sure we touch on this before releasing, as we had a similar solution in v3 that ultimately caused the bundler to lose funds. Here are some relevant threads
> 
> * https://biconomyworkspace.slack.com/archives/C066X6E3J9Y/p1704888602987659
> * https://biconomyworkspace.slack.com/archives/C03G74F8W76/p1703835030390559

I don't see how this implementation could be related to that, we are getting the gas values dinamically here, nothing is hardcoded or I didn't understand the real issue on that thread.

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -499,11 +507,75 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+        if (paymasterServiceData.mode === PaymasterMode.SPONSORED) {
+          const paymasterData = await (this.paymaster as IHybridPaymaster<PaymasterUserOperationDto>).getPaymasterAndData(userOp, {
+            mode: paymasterServiceData.mode,
+          });
+          userOp.paymasterAndData = paymasterData.paymasterAndData;
+          userOp.callGasLimit = paymasterData.callGasLimit;
+          userOp.verificationGasLimit = paymasterData.verificationGasLimit;
+          userOp.preVerificationGas = paymasterData.preVerificationGas;
+          return userOp;
+        } else if (paymasterServiceData.mode === PaymasterMode.ERC20) {
```
livingrockrises, line 532: if the mode is ERC20 and if a preferred token in given in paymasterServiceData then we keep this part here only for this specific case. Downside still is that user won't be able to see expected fee quote if it's part of ux flow.

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -499,11 +507,75 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+          const finalUserOp = await this.buildTokenPaymasterUserOp(userOp, {
+            // @ts-expect-error There should always be a fee quote
+            feeQuote: feeQuotesResponse.feeQuotes[0]!,
+            spender: (feeQuotesResponse.tokenPaymasterAddress as Hex) || "",
+            maxApproval: true,
+          });
+          const newPaymasterServiceData = {
+            mode: PaymasterMode.ERC20,
+            // @ts-expect-error There should always be a fee quote tokenAddress
+            feeTokenAddress: feeQuotesResponse.feeQuotes[0].tokenAddress!, // TODO: check if this is correct
```
joepegler, line 549: Can we throw if that's not the case with a relevant message?

packages/account/src/BiconomySmartAccountV2.ts
joepegler: I don't think you'll have to change the return type if you throw an error. Unless there's a good reason not to I would throw here, otherwise users won't know if something has failed and proceed assuming they have gotten the correct gas values. Wouldn't that be exactly the problem described in this thread? https://biconomyworkspace.slack.com/archives/C05E8EYSJ3E/p1703605643390259 when we provide incorrect gas values during sendUserOp our bundler is currently (incorrectly) setup in a way which loses money, so we should error if we can't find them. I would remove the try catch here because of this. Please someone cmiiw

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -499,11 +500,66 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+      return userOp;
+    }
+  }
+
   /**
    *
    * @param userOp
    * @param params
-   * @description This function call will take 'unsignedUserOp' as an input, sign it with the owner key, and send it to the bundler.
+   * @description This function call will setup paymaster values (if paymaster is used), take 'unsignedUserOp' as an input, sign it with the owner key, and send it to the bundler.
```
livingrockrises, line 562: the implementation here doesn't look like it will setup paymaster values

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -640,7 +698,7 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
    * @param buildUseropDto options for building the userOp
    * @returns Promise<UserOpResponse>
    */
-  async sendTransaction(manyOrOneTransactions: Transaction | Transaction[], buildUseropDto?: BuildUserOpOptions) {
+  async sendTransaction(manyOrOneTransactions: Transaction | Transaction[], buildUseropDto?: BuildUserOpOptions): Promise<UserOpResponse> {
```
livingrockrises, line 701: just adding a note here,

1. options to do either from below,
dummy pnd this.getDummyPnd(). paymaster.getdummypnD(). override paymasterAndData

2. making use of supplied overrides
only for bundler estimations (can be additional param in bundler.estimateUserOpGas without overriding userOp)

potentially adding a new method buildSponsoredUserOpOptimised() which will rely on calling paymaster only for gas estimations and expects paymasterAndData return.

livingrockrises (commented):
review i

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45: This is for next iteration yes ?

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45: Fixed

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45: This was fixed

packages/account/src/BiconomySmartAccountV2.ts
livingrockrises: yes we can just throw. in buildUserOp we can catch this and rethrow. If we want to handle differently later with strictMode it can be easily done imo

packages/account/src/BiconomySmartAccountV2.ts
livingrockrises: yes can be separate PR

packages/account/tests/account.e2e.spec.ts
joepegler: const balance = (await checkBalance(publicClient, await smartWallet.getAddress() )) as bigint;

packages/account/tests/account.e2e.spec.ts
joepegler: Did it work? This will end up in all our documentation if it doesn't. It's worth testing

packages/account/tests/account.e2e.spec.ts
joepegler: Should work for nativeToken, though I've not tested it

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -499,11 +500,64 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+   * @param {Partial<UserOperationStruct>} userOp - The partial user operation structure to be modified.
+   * @param {PaymasterUserOperationDto} paymasterServiceData - The paymaster service data containing mode and additional information.
+   * @returns {Promise<Partial<UserOperationStruct>>} A promise that resolves to the modified user operation structure.
+   */
+  async setPaymasterUserOp(
+    userOp: Partial<UserOperationStruct>,
+    paymasterServiceData: PaymasterUserOperationDto,
+  ): Promise<Partial<UserOperationStruct>> {
+    if (this.paymaster !== undefined) {
+      try {
```
joepegler, line 515: I would remove the try catch: This loses the important information inside 'e':
```
try {...}
catch(e) {
  throw ("Some generic error");
}
```

joepegler (changes requested):

AmanRaj1608 (approved):

livingrockrises (approved):

packages/account/src/utils/Types.ts
```diff
@@ -177,6 +177,20 @@ export type InitilizationData = {
   signerAddress?: string;
 };
 
+export type PaymasterUserOperationDto = {
+  mode: PaymasterMode;
+  calculateGasLimits?: boolean;
+  expiryDuration?: number;
+  webhookData?: {
```
joepegler, line 184: Record<string, any>

joepegler (approved):
LGTM, nice job

joepegler:
Can you remove all the relevant code in the sdk-examples project when you can? 
https://github.com/bcnmy/sdk-examples/blob/474353ad25e993ce4b11d4e65d8b9ee13a101f2c/scripts/gasless/mintNftEthers.ts#L54

VGabriel45:
> Can you remove all the relevant code in the sdk-examples project when you can? https://github.com/bcnmy/sdk-examples/blob/474353ad25e993ce4b11d4e65d8b9ee13a101f2c/scripts/gasless/mintNftEthers.ts#L54

Yea, I will update sdk examples "backend-node" project and test it with v4. After that I will push a new branch and we will merge it after publishing.

Alximm (approved):

Alximm (approved):

Issue #373, opened 2024-01-18
`UserOpResponse.wait` does not wait for enough confirmations


au2001:
## Problem You're Facing

After sending a userOp, I want to wait for N confirmations on the blockchain.\
Right now, using `userOpResponse.wait(999)` returns a Promise which resolves right after the transaction is included in a block, without waiting for any additional confirmation.

## Proposed Solution

The `UserOpResponse.wait` function already takes as argument the number of desired confirmations.\
That function should simply be fixed to correctly take into account this value.

## Alternatives Considered

A possible workaround is to use `waitForTxHash` and wait for the desired number of confirmations for that transaction using another library (e.g. ethers).

## Use Cases

Waiting for a given number of confirmations can help prevent issues with blockchain forks/reorgs.

## Additional Info

I'm willing to provide a fix and will open a PR which closes this issue.

livingrockrises:
we will look into this and also review the PR

PR #374, opened 2024-01-18
Fix UserOpResponse wait confirmations

au2001:
# Summary

The `wait` and `waitForTxHash` functions of the `UserOpResponse` type did not work as expected.\
This PR addresses some issues including waiting for a given number of confirmations, and incorrectly thrown errors after the Promise resolved.

Related Issue: #373

## Change Type

- [x] Bug Fix
- [ ] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

- [x] I've added comments for any hard-to-understand areas (N/A)
- [ ] I've added tests that prove my fix is effective or my feature works*

# Additional Information

This bug fix is technically a breaking change, as code which uses `userOpResponse.wait(1)` will change behavior.\
It previously would not wait for the confirmation, so this PR will approximately add an N block delay to all `.wait(N)` calls.

There can still be race conditions in both `wait` and `waitForTxHash` if the RPC is sometimes slow and takes more than `intervalValue` milliseconds to answer. This is not covered by this PR and is negligible IMHO.

\* Tests have not been implemented because of [this comment](https://github.com/bcnmy/biconomy-client-sdk/blob/9d7f52dc06d20cefd0a0074bad061f29e3895fd1/packages/account/tests/SmartAccountV2.local.spec.ts#L428-L433).


packages/bundler/src/Bundler.ts
```diff
@@ -156,14 +156,18 @@ export class Bundler implements IBundler {
                   if (confirmations >= confirmedBlocks) {
                     clearInterval(intervalId);
                     resolve(userOpResponse);
+                    return;
```
livingrockrises, line 159: Omitting the return statement in this case is common and doesn't affect the functionality of the code as it doesn't make practical difference.

packages/bundler/src/Bundler.ts
```diff
@@ -156,14 +156,18 @@ export class Bundler implements IBundler {
...
+                  resolve(userOpResponse);
+                  return;
                 }
-                clearInterval(intervalId);
-                resolve(userOpResponse);
               }
             } catch (error) {
               clearInterval(intervalId);
               reject(error);
+              return;
```
livingrockrises, line 170: same here

packages/bundler/src/Bundler.ts
```diff
@@ -156,14 +156,18 @@ export class Bundler implements IBundler {
...
                     clearInterval(intervalId);
                     resolve(userOpResponse);
+                    return;
                   }
+                } else {
+                  clearInterval(intervalId);
+                  resolve(userOpResponse);
+                  return;
                 }
-                clearInterval(intervalId);
```
livingrockrises, line 161: this is practically already in else

packages/bundler/src/Bundler.ts
```diff
@@ -186,18 +190,19 @@ export class Bundler implements IBundler {
...
-          const intervalId = setInterval(() => {
-            this.getUserOpStatus(sendUserOperationResponse.result)
-              .then((userOpStatus) => {
-                if (userOpStatus && userOpStatus.state && userOpStatus.transactionHash) {
-                  clearInterval(intervalId);
-                  resolve(userOpStatus);
-                }
-              })
-              .catch((error) => {
+          const intervalId = setInterval(async () => {
```
livingrockrises, line 193: This changes also doesn't make any difference other than using async instead of promise then catch pattern.

packages/bundler/src/Bundler.ts
au2001: Issue #374 arises precisely because this is not in an else block: if the `confirmations >= confirmedBlocks` condition is false, it should not `resolve`.

packages/bundler/src/Bundler.ts
au2001: You're correct on the fact that this is simply changing `.then`/`.catch` to `async`/`await`.
But it does change the behavior, because the timeout check below was not within the `.then` block.
So now, the timeout check will only run after the actual verification of the status of the user op has finished.
This change was made for this function to be consistent with the other `wait` function.

packages/bundler/src/Bundler.ts
au2001: True, in the current state it does not make any practical difference.
I added it to skip the timeout check below which is not needed, to save a few CPU cycles.

The [`Promise` docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/Promise) cite that calling `reject` after `resolve` does not have any effect, and [`clearInterval`](https://developer.mozilla.org/en-US/docs/Web/API/clearInterval) is idempotent, so this `return` statement can safely be removed.
But, in case side-effects are added below, and for performance's sake, it is cleaner IMHO to keep this `return` statement.

If it's still your opinion that I should remove it though, I will do so.

livingrockrises: I see!

packages/bundler/src/Bundler.ts
livingrockrises: Hmm It makes sense that the timeout check is conditioned on the completion of the getUserOpStatus operation, making the behavior more consistent with the outcome of that operation. 

But also in this case, imo it should not be a problem even if it would run independently of the resolution or rejection of the promise returned by getUserOpStatus.

However it improves readability and makes consistent

livingrockrises (commented):
Thanks for the fix. reviewed!

livingrockrises:
Could you also change the base to develop please

packages/bundler/src/Bundler.ts
au2001: Absolutely, that's an oversight on my end. I've changed it now.

au2001:
> @livingrockrises Could you also change the base to develop please

Done!

livingrockrises (approved):
LGTM

AmanRaj1608 (approved):

PR #375, opened 2024-01-18
Apply Signer Abstraction to modules

joepegler:
# Summary

A devEx improvement that applies the same signer abstraction to the relevant modules, so they no longer require Alchemys Signer when instantiated.

Previously:

```
  import { WalletClientSigner } from "@alchemy/aa-core";
  import { MultiChainValidationModule } from "@biconomy/modules";
  const signer = new WalletClientSigner(viemWallet, "viem");
  const multiChainModule = await MultiChainValidationModule.create({
    signer: signer,
    moduleAddress: DEFAULT_MULTICHAIN_MODULE,
  });
  const biconomySmartAccount1 = await createSmartWalletClient({
    chainId: config.chainId,
    rpcUrl: config.rpcUrl,
    signer: signer,
    bundlerUrl: config.bundlerUrl,
    biconomyPaymasterApiKey: config.biconomyPaymasterApiKey,
    defaultValidationModule: multiChainModule,
    activeValidationModule: multiChainModule,
  });
```

Now:
```
  import { createSmartWalletClient } from "@biconomy/account";
  import { createMultiChainValidationModule } from "@biconomy/modules";
  const defaultValidationModule = await createMultiChainValidationModule({ signer });
  const smartWallet = await createSmartWalletClient({ bundlerUrl, defaultValidationModule });
```

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [ ] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [x] Other

# Checklist

# Additional Information

# Branch Naming


packages/common/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@biconomy/common",
+  "version": "3.1.2",
```
livingrockrises, line 3: shouldn't this version be same as other package versions?

livingrockrises: ok just saw we can update them all together to 3.1.3 or 4.0.0 for now. 
because common 3.1.2 already exists in npm registry

packages/common/src/utils/Helpers/convertSigner.ts
```diff
@@ -0,0 +1,57 @@
+import { EthersSigner } from "../EthersSigner";
+import { SupportedSigner } from "../Types";
+import { WalletClient } from "viem";
+import { WalletClientSigner, SmartAccountSigner } from "@alchemy/aa-core";
```
livingrockrises, line 4: what's stopping up from not importing stuff from alchemy/aa-core?
cc @AmanRaj1608

joepegler: I don't follow

joepegler: image
Our main BiconomySmartAccountV2 class extends it

livingrockrises: Hmm I was thinking to create a local copy of this so every time we are not forced to update and wait for alchemy to make some changes.

livingrockrises (approved):

AmanRaj1608 (approved):

AmanRaj1608:
Now that I think about, we should not another lib common. Instead just reuse the same file in utils in module and account package. @joepegler 

Otherwise we need to publish another common package, which  is not needed as of now much

But if we move to this common package flow of should also do:
- Move Logger to common
- Move HttpRequest to common

joepegler:
Can do that no problem, it's not a big change. So the modules package should import convertSigner from the accounts package instead? Should all common helpers essentially be in our accounts package too then? We do have quite a bit of duplicated code. I can tidy that too

PR #388, opened 2024-01-25
Feat(SMA-559): Improved dx for erc20 paymaster calls

joepegler:
# Summary

The devEx for erc20 gas payments was difficult to follow I have simplified the flow for consumers so that erc20 paymaster such that it can be used in the following flow:

```
const fees = await smartWallet.getTokenFees(tx, ...);
const feeData = fees[0]; // Select the fee.
const result = await sendTransaction(tx, { ...paymasterDetails, ...feeData });
```

Related Issue: SMA-559

## Change Type

- [ ] Bug Fix
- [x] Refactor
- [x] New Feature
- [ ] Breaking Change
- [x] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist


packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -502,16 +503,59 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
     return encodeAbiParameters(parseAbiParameters("bytes, address"), [moduleSignature, moduleAddressToUse]);
   }
 
+  async #getPaymasterUserOp(
```
VGabriel45, line 506: What's the point of "#" in front of these ? For "internal" functions ?

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -502,16 +503,59 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+        }
+        return this.#getPaymasterAndDataFromPaymaster(userOp, { ...paymasterServiceData, feeQuote, feeTokenAddress: preferredToken });
+      } else {
+        throw new Error("FeeQuote was not provided, please call smartAccount.getTokenFees() to get feeQuote");
+      }
+    }
+    throw new Error("Invalid paymaster mode");
+  }
+
+  async #getPaymasterAndDataFromPaymaster(
```
VGabriel45, line 538: Why "getPaymasterAndDataFromPaymaster" and not "getPaymasterAndData" ? Is this because you want to mention that the gas estimations are coming from paymaster and not bundler ?

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -502,16 +503,59 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+    userOp: Partial<UserOperationStruct>,
+    paymasterServiceData: PaymasterUserOperationDto,
+  ): Promise<Partial<UserOperationStruct>> {
+    const paymaster = this.paymaster as IHybridPaymaster<PaymasterUserOperationDto>;
+    const paymasterData = await paymaster.getPaymasterAndData(userOp, paymasterServiceData);
+    return { ...userOp, ...paymasterData };
+  }
+
+  async #getTokenFeesFromUserOp(userOp: Partial<UserOperationStruct>, feeQuotesOrData: FeeQuotesOrDataDto): Promise<FeeQuotesOrDataResponse> {
```
VGabriel45, line 547: I would rather name this "getTokenFeesForUserOp" because you're not reading the token fees from the user op but rather for the user op. What do you think ?

packages/account/src/BiconomySmartAccountV2.ts
joepegler: Yeah they denote private class properties

packages/account/src/BiconomySmartAccountV2.ts
joepegler: Yes that's right. I thought getPaymasterAndData might be confusing on the SmartWallet class considering that's the name of the function on the paymaster, but I guess it's not that confusing. Will happily change it back

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -502,16 +503,59 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
     return encodeAbiParameters(parseAbiParameters("bytes, address"), [moduleSignature, moduleAddressToUse]);
   }
 
+  async #getPaymasterUserOp(
+    userOp: Partial<UserOperationStruct>,
+    paymasterServiceData: PaymasterUserOperationDto,
+  ): Promise<Partial<UserOperationStruct>> {
+    if (paymasterServiceData.mode === PaymasterMode.SPONSORED) {
+      return this.#getPaymasterAndDataFromPaymaster(userOp, paymasterServiceData);
+    } else if (paymasterServiceData.mode === PaymasterMode.ERC20) {
```
VGabriel45, line 512: The "buildTokenPaymasterUserOp" is not called anymore ?

packages/account/src/BiconomySmartAccountV2.ts
joepegler: Yeah np, can do

packages/account/src/BiconomySmartAccountV2.ts
joepegler: Oh Christ, nice catch. I wonder how tests didn't fail 🤔

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45: Can we stick do the current implementation of using "private" for private functions ? Or is there any reason using # would be better ?

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45: I think "getPaymasterAndData" already speaks that is calling paymaster

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45: Tests did not fail because "buildTokenPaymasterUserOp" adds token approval user op in order to approve the paymaster to spend the ERC20 for the user but probably our account that we use for testing already gave max approval to the paymaster.

joepegler: Nice catch. Could have been a nasty bug

joepegler: Updated

VGabriel45 (approved):

livingrockrises:
const fees = await smartWallet.getTokenFees(tx, ...);

so this will call buildUserOp first and then call getFeeQuotesOrData ?

joepegler:
> const fees = await smartWallet.getTokenFees(tx, ...);
> 
> so this will call buildUserOp first and then call getFeeQuotesOrData ?

Yes. This isn't an optimisation, it's only to fix the poor devEx. We can work on optimisations in the next version

livingrockrises (approved):

PR #390, opened 2024-01-26
Added simulationType flag

VGabriel45:
# Summary

Added simulationType flag in sendSignedUserOp() and in buildUseropDto type from sendTransaction()

## Change Type

- [ ] Bug Fix
- [x] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

### DIFF
--- a/packages/bundler/src/Bundler.ts
+++ b/packages/bundler/src/Bundler.ts
@@ -27,7 +27,7 @@ import {
   DEFAULT_ENTRYPOINT_ADDRESS,
 } from "./utils/Constants";
 import { sendRequest, HttpMethod } from "@biconomy/common";
-import { extractChainIdFromBundlerUrl } from "utils/Utils";
+import { extractChainIdFromBundlerUrl } from "./utils/Utils";
 
 /**
  * This class implements IBundler interface.


packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -670,17 +671,18 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
   /**
    *
-   * @param userOp
+   * @param userOp - The signed user operation to send
+   * @param simulationType - The type of simulation to perform ("validation" | "validation_and_execution")
    * @description This function call will take 'signedUserOp' as input and send it to the bundler
    * @returns
    */
-  async sendSignedUserOp(userOp: UserOperationStruct): Promise<UserOpResponse> {
+  async sendSignedUserOp(userOp: UserOperationStruct, simulationType?: SimulationType): Promise<UserOpResponse> {
```
livingrockrises, line 685: this would probably conflict with params once we add state overrides stuff. it's probably better to keep separate type SendSignedUserOpParams which is passed a second argument.

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -697,7 +699,7 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
     this.validateUserOp(userOp, requiredFields);
     if (!this.bundler) throw new Error("Bundler is not provided");
     Logger.warn("userOp being sent to the bundler", userOp);
-    const bundlerResponse = await this.bundler.sendUserOp(userOp);
+    const bundlerResponse = await this.bundler.sendUserOp(userOp, simulationType);
```
livingrockrises, line 702: same here on the bundler. it's like EthSendUserOperationParams

packages/account/src/utils/Types.ts
```diff
@@ -142,6 +142,7 @@ export type BuildUserOpOptions = {
   nonceOptions?: NonceOptions;
   forceEncodeForBatch?: boolean;
   paymasterServiceData?: PaymasterUserOperationDto;
+  simulationType?: SimulationType;
```
livingrockrises, line 145: buildUserOp only calls eth_estimateUserOpGas on the bundler. so we probably should not have this here..
if smartWallet.sendTransaction calls buildUserOp and then sendUserop / sendSignedUserOp this only needs to be added in sendTransaction params as dto.

packages/account/src/BiconomySmartAccountV2.ts
livingrockrises: once we have that object, in another PR we can add overrides in same dto.

so simulationType is only relevant for biconomy bundler
but stateOverrideSet may be accepted by one or more bundlers 
(will have to mention this in comments)

livingrockrises (commented):
review i

packages/account/src/utils/Types.ts
VGabriel45: Yea, will remove it and add it as a separate optional param on sendTransaction or I can create a dto and use that as a param on sendTransaction which would look better but will change the way users will need to provide params. It's better to not add this in v4

packages/account/src/BiconomySmartAccountV2.ts
VGabriel45: Why would it conflict ? state overrides is a param that will be passed to buildUseropDto, as it can be used only on estimate user op gas so that param will not be on sendSignedUserOp

livingrockrises (approved):
LGTM

PR #424, opened 2024-02-20
Feat(SMA-598): Smart Account Withdrawals

joepegler:
# Summary

Public utility method which withdraws funds from Smart Account to recipient (usually EOA)

Related Issue: SMA-598

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

# Additional Information

# Branch Naming


packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -257,13 +258,105 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+   * // OR to withdraw all of the native token, leaving no dust in the smart account
+   *
+   * const { wait } = await smartAccount.withdraw(account.pubKey, [], {
+   *  paymasterServiceData: { mode: PaymasterMode.SPONSORED },
+   * });
+   *
+   * const { success } = await wait();
+   */
+  public async withdraw(
+    recipient: Hex,
```
livingrockrises, line 304: we could also make recipient per each withdraw request

livingrockrises: optionally. if not provided use the main recipient given above

packages/account/src/BiconomySmartAccountV2.ts
```diff
@@ -257,13 +258,105 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+   *   chain: polygonMumbai,
+   *   transport: http(),
+   * });
+   *
+   * const smartAccount = await createSmartAccountClient({ signer, bundlerUrl, biconomyPaymasterApiKey });
+   *
+   * const { wait } = await smartAccount.withdraw(
+   *  account.pubKey, // recipient
+   *  [
+   *    { address: USDT, amount: BigInt(1) },
```
livingrockrises, line 287: Are they supposed to give wei value? 
or just like symbol 'USDT' and amount 0.5 is possible?

livingrockrises: we'd have to keep symbols and chain and token decimals mapping.
and then find out using current chainId..

packages/account/src/utils/Types.ts
```diff
@@ -39,6 +39,13 @@ export interface BalancePayload {
   formattedAmount: string;
 }
 
+export interface WithdrawalRequest {
+  /** The address of the tokenAddress */
+  address: Hex;
```
livingrockrises, line 44: could name destination or recipient or receiver

packages/account/tests/account.e2e.spec.ts
```diff
@@ -477,4 +476,246 @@ describe("Account Tests", () => {
...
+      signer,
+      biconomyPaymasterApiKey,
+      bundlerUrl,
+    });
+
+    const smartAccountAddress = await smartAccount.getAddress();
+    const balanceOfSABefore = (await checkBalance(publicClient, smartAccountAddress)) as bigint;
+    const balanceOfRecipientBefore = (await checkBalance(publicClient, smartAccountOwner)) as bigint;
+
+    const { wait } = await smartAccount.withdraw(smartAccountOwner, undefined /* null or undefined or [] */, {
```
livingrockrises, line 681: how do you choose the amount to withdraw if paymaster is not involved?

if a paymaster call fails, we would withdraw all which will fail on chain again

packages/account/src/BiconomySmartAccountV2.ts
joepegler: No I'm not using symbols here, it's a can of worms (we'd need a database of ticker mappings, and it'd have to be 100% reliable) as there's no easy way to match a symbol to an address. Happy to revisit this if people ask for it though. As it is here the dev will need to know their own addresses - I think that's fine.

joepegler: Yes they give gwei value. In the [tsdoc here](https://github.com/bcnmy/biconomy-client-sdk/pull/424/files#diff-77f53b7690a5e2bc101257d2fb7aab39eef4cc93964704ecb30e8b8133ec81d6R45) I've made this clear. I also think this is what devs will expect. Doing the opposite is more likely to cause confusion

packages/account/src/BiconomySmartAccountV2.ts
joepegler: They can always call withdraw multiple times too per recipient. Actually having a method named 'withdraw' that doesn't go to the owner is slightly confusing, but I couldn't think of a clear path to finding the recipient without it being provided by the dev. I would have hardcoded this if I could have... 

Can you think of a way to reliably find the owner in this case? I think it would be best now that I'm thinking about it...

packages/account/src/utils/Types.ts
joepegler: No this is the asset address, not the recipient.

packages/account/tests/account.e2e.spec.ts
joepegler: They could call the getBalances method :)

joepegler: Can you elaborate on this second point? I don't follow

livingrockrises (commented):
review i

packages/account/tests/account.e2e.spec.ts
livingrockrises: if I want to withdraw native tokens and max amount (say 1 Matic) what's there in SA 
and if there is no paymaster or paymaster call fails then pnd goes '0x' which means account payas for gas in native tokens. so we can't transfer 1 and also pay 0.01 matic in gas

packages/account/src/BiconomySmartAccountV2.ts
livingrockrises: hmm thinking. we could have method single withdraw (which also acts as withdrawERC20) 
and also withdrawMultipleERC20 and give them an option to provide receiver for each withdrawal. 

these methods are risky though cause if we make mistake in sdk or choose to go rough then all tokens would be transferred to us

packages/account/src/BiconomySmartAccountV2.ts
livingrockrises: got it

packages/account/src/BiconomySmartAccountV2.ts
livingrockrises: I mean
```
[
   *    { address: USDT, amount: BigInt(1) }
```

here they could also provide recipient address, instead of having common recipient..or maybe if not provided then fallback to common recipient given above. 

I agree with you it could by default go to the owner in most cases where owner is another hot wallet metamask EOA. but they might want to send elsewhere or to a friend.

livingrockrises (approved):
review ii

packages/account/tests/account.e2e.spec.ts
joepegler: Yes that's right. I have thrown [this error](https://github.com/bcnmy/biconomy-client-sdk/pull/424/files/fecb066bce1f444f4537e5bdf518ceddac40d2d2#diff-78e7afa7e87e267fae298f597794c5ab45544d36cd675d4dd9aa843337bb652bR337) if a user does not provide his own amount in this case. dev will be expected to find his own amounts to withdraw if no paymaster is used. It'll be on the dev to find the right amount

packages/account/src/BiconomySmartAccountV2.ts
joepegler: OK makes sense. Will add.

packages/account/tests/account.e2e.spec.ts
joepegler: https://github.com/bcnmy/biconomy-client-sdk/pull/424/files#diff-78e7afa7e87e267fae298f597794c5ab45544d36cd675d4dd9aa843337bb652bR349

joepegler: I'm basically telling the dev that it's up to them to pick the amount of dust they want to leave behind, assuming they're not using a paymaster.

VGabriel45 (approved):

AmanRaj1608 (approved):

livingrockrises:
these are all new methods so let's send it and get more feedback.

PR #426, opened 2024-02-20
Feat(SMA-663): Minimal Reproducible Example request

joepegler:
Updated the bug report guideline to more forcibly request a "Minimal Reproducible Example"


packages/account/src/index.ts
```diff
@@ -16,7 +16,7 @@ export {
   type FeeQuotesOrDataResponse,
   createPaymaster,
 } from "@biconomy/paymaster";
-export { EthersSigner, convertSigner, type LightSigner } from "@biconomy/common";
+export { EthersSigner, convertSigner, type LightSigner, type SupportedSigner } from "@biconomy/common";
```
livingrockrises, line 19: what is this change for? does it have to be part of this PR purpose?

livingrockrises (approved):
reviewed

packages/account/src/index.ts
joepegler: No it's not relevant to this PR. I noticed we'll need to export this if the version of viem that we bundle gets dated - users will need to cast their signer. The long term solution for this is to move viem to a peer dep.

livingrockrises:
conflict with lockfile

joepegler:
fixed conflicts

VGabriel45 (approved):

PR #445, opened 2024-03-04
Improved balances

joepegler:
# Summary

Minor improvement around the dx of the getBalance call. It will now always return the native token balance in the response.


livingrockrises (approved):

livingrockrises:
conflict to be resolved

VGabriel45 (approved):

PR #451, opened 2024-03-07
Fix: eslint file extension check 

joepegler:


VGabriel45 (approved):

packages/account/src/utils/index.ts
```diff
@@ -1,3 +1,3 @@
-export * from "./Types";
-export * from "./Utils";
-export * from "./Constants";
+export * from "./Types.js";
+export * from "./Utils.js";
+export * from "./Constants.js";
```
AmanRaj1608, line 3: Why is this needed

AmanRaj1608 (approved):

packages/account/src/utils/index.ts
joepegler: For esm builds. https://stackoverflow.com/questions/62619058/appending-js-extension-on-relative-import-statements-during-typescript-compilat/62626938#62626938

joepegler: In the new SDK we have a more robust solution: tsc-alias
https://github.com/bcnmy/sdk/blob/main/package.json#L25

AmanRaj1608 (approved):

PR #458, opened 2024-04-04
feat: added support for PrivateKeyAccount as signer in account config

VGabriel45:
# Summary

Added PrivateKeyAccount support as direct signer in account config.

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

### DIFF
--- a/packages/account/src/utils/Utils.ts
+++ b/packages/account/src/utils/Utils.ts
@@ -55,7 +55,7 @@ export const compareChainIds = async (
   biconomySmartAccountConfig: BiconomySmartAccountV2Config,
   skipChainIdCalls: boolean,
 ): Promise<Error | void> => {
-  const signerResult = await convertSigner(signer, skipChainIdCalls);
+  const signerResult = await convertSigner(signer, skipChainIdCalls, biconomySmartAccountConfig.rpcUrl);
 
   const chainIdFromBundler = biconomySmartAccountConfig.bundlerUrl
     ? extractChainIdFromBundlerUrl(biconomySmartAccountConfig.bundlerUrl)
--- a/packages/account/tests/account.spec.ts
+++ b/packages/account/tests/account.spec.ts
@@ -1,4 +1,4 @@
-import { Paymaster, createBundler, createSmartAccountClient } from "../src";
+import { ERROR_MESSAGES, Paymaster, createBundler, createSmartAccountClient } from "../src";
 import { Chain, createWalletClient, http } from "viem";
 import { localhost } from "viem/chains";
 import { privateKeyToAccount, generatePrivateKey } from "viem/accounts";
@@ -307,6 +307,6 @@ describe("Account Tests", () => {
       bundlerUrl,
     });
 
-    await expect(createSmartAccount).rejects.toThrow("rpcUrl is required for PrivateKeyAccount signer type, please provide it in the config");
+    await expect(createSmartAccount).rejects.toThrow(ERROR_MESSAGES.MISSING_RPC_URL);
   }, 50000);
 });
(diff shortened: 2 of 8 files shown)


packages/account/tests/account.spec.ts
```diff
@@ -281,4 +280,33 @@ describe("Account Tests", () => {
...
+    const { bundlerUrl } = ganache;
+
+    const account = privateKeyToAccount(`0x${process.env.E2E_PRIVATE_KEY_ONE}`);
+
+    const createSmartAccount = createSmartAccountClient({
+      signer: account,
+      bundlerUrl,
+    });
+
+    await expect(createSmartAccount).rejects.toThrow("rpcUrl is required for PrivateKeyAccount signer type, please provide it in the config");
```
joepegler, line 310: Can you use the ERRORS config for this?

packages/common/src/utils/Helpers/convertSigner.ts
```diff
@@ -51,7 +50,17 @@ export const convertSigner = async (signer: SupportedSigner, skipChainIdCalls: b
       }
       // convert viems walletClient to alchemy's SmartAccountSigner under the hood
       resolvedSmartAccountSigner = new WalletClientSigner(walletClient, "viem");
-      rpcUrl = walletClient?.transport?.url ?? null;
+      rpcUrl = walletClient?.transport?.url ?? undefined;
+    } else if (signer as PrivateKeyAccount) {
```
joepegler, line 54: I think this should be `instanceof` instead of `as`

joepegler (changes requested):

packages/common/src/utils/Helpers/convertSigner.ts
joepegler: You're casting every other signer to a PrivateKeyAccount here, - not just accounts.

packages/account/src/utils/Utils.ts
```diff
@@ -55,7 +55,7 @@ export const compareChainIds = async (
   biconomySmartAccountConfig: BiconomySmartAccountV2Config,
   skipChainIdCalls: boolean,
 ): Promise<Error | void> => {
-  const signerResult = await convertSigner(signer, skipChainIdCalls);
+  const signerResult = await convertSigner(signer, skipChainIdCalls, biconomySmartAccountConfig.rpcUrl);
```
livingrockrises, line 58: why is this added?

livingrockrises (commented):
review i

packages/common/src/utils/Helpers/convertSigner.ts
VGabriel45: instanceof works with classes, it does not work with types

VGabriel45: > You're casting every other signer to a PrivateKeyAccount here, - not just accounts.

Not sure what do you mean as this is the last "else if", it will cast it only if signer is indeed of PrivateKeyAccount type, otherwise it will break and throw an error.

packages/account/src/utils/Utils.ts
VGabriel45: If a user passes a PrivateKeyAccount there is no info about transport so he needs to at least pass rpcUrl in the config so we can properly construct the walletClient with the correct transport

livingrockrises (approved):

packages/common/src/utils/Helpers/convertSigner.ts
VGabriel45: This was changed, please review

joepegler:
LGTM. We can close this as I have added it here: https://github.com/bcnmy/biconomy-client-sdk/pull/463/commits/bb9ba8edfba18b667154309bc3098232b7bd79e9

PR #470, opened 2024-04-19
added fetch for token balance in getSupportedTokens

VGabriel45:
# Summary

Added a fetch for token balance in getSupportedTokens method which is returning supported erc20 tokens for paymaster payments. This method output will now also contain token balance of the smart account.

Created a new method **getTokenBalance()** which is similar to **getTokenBalanceInfo()** (which has been renamed from **getBalance()**), the scope of this new method is to fetch only the token balance and nothing more.

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

### DIFF
--- a/src/account/BiconomySmartAccountV2.ts
+++ b/src/account/BiconomySmartAccountV2.ts
@@ -345,9 +345,9 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
   }
 
   /**
-   * Returns token balances (and native token balance) of the smartAccount instance.
+   * Returns tokens balance and information of the tokens (for native and erc20 tokens) for the smartAccount instance.
    *
-   * This method will fetch the token balances of the smartAccount instance.
+   * This method will fetch tokens info given an array of token addresses for the smartAccount instance.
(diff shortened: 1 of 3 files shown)


src/account/BiconomySmartAccountV2.ts
```diff
@@ -442,6 +442,49 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+   *
+   * const token = "0x747A4168DB14F57871fa8cda8B5455D8C2a8e90a";
+   * const smartAccount = await createSmartAccountClient({ signer, bundlerUrl });
+   * const tokenBalance = await smartAccount.getTokenBalance(token);
+   *
+   * console.log(tokenBalance);
+   * // 83546776652078153068n
+   *
+   */
+  public async getTokenBalance(address?: Hex): Promise<bigint> {
```
joepegler, line 473: this function returns a users ethBalance, so the name is now less relevant. `getBalance` seems more appropriate

joepegler (changes requested):

joepegler (approved):
lgtm, nice

livingrockrises (approved):

PR #474, opened 2024-04-23
feat: gas offset param for sendTransaction and sendUserOp

VGabriel45:
# Summary

Implemented a gas offset parameter for utilization with sendTransaction & buildUserOp + sendUserOp.

This additional parameter enables the augmentation of gas values previously estimated through Biconomy's infrastructure.
When employing a paymaster and specifying a gas offset, calculateGasLimits will be disabled to prevent the paymaster's gas estimations from overriding your offset.

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

- [] My code follows this project's style guidelines

---

## PR-Codex overview
This PR updates package configurations and refactors functions related to gas calculations and user operations.

### Detailed summary
- Updated `package.json` files configuration
- Refactored gas-related functions for user operations
- Added new gas offset percentage types for calculations
- Modified gas limit values in `BiconomyPaymaster.ts`
- Added new functions for gas calculations in `Utils.ts`
- Updated `Types.ts` with new options for building user operations

> The following files were skipped due to too many changes: `src/account/BiconomySmartAccountV2.ts`, `tests/account/write.test.ts`


src/account/utils/Types.ts
```diff
@@ -178,26 +178,28 @@ export type BiconomySmartAccountV2ConfigConstructorProps =
...
-  /**  paymasterServiceData: Options specific to transactions that involve a paymaster */
-  paymasterServiceData?: PaymasterUserOperationDto
-  /**  simulationType: Determine which parts of the tx a bundler will simulate: "validation" | "validation_and_execution".  */
-  simulationType?: SimulationType
-  /**  stateOverrideSet: For overriding the state */
-  stateOverrideSet?: StateOverrideSet
-  /** set to true if the tx is being used *only* to deploy the smartContract, so "0x" is set as the userOp.callData  */
-  useEmptyDeployCallData?: boolean
-}
+  gasOffset?: GasOffset;
```
joepegler, line 194: Instead of it being an amount that you increment as an absolute value, could it not be expressed as a percentage? Like I would like to be able to increase the amount be 10% by providing an 'offsetFactor' of '+0.1', where the original estimate gets multiplied by `1.1`. I think this might be more useful

joepegler (commented):
V nice. Had one point regarding using an 'offsetFactor' instead of 'offsetIncrement'. Any thoughts?

src/account/utils/Types.ts
VGabriel45: Let me check how this would be implemented, if it adds too much complexity I don't think it's worth it

joepegler: We could even hardcode the offsetFactors: `1`, `1.2` & `1.4` and then offer them as "SLOW", "MEDIUM" and "FAST" for the sake of familiarity of MM users

VGabriel45: Interesting, like "slippage" in DeFi, will look into this

VGabriel45: "SLOW", "MEDIUM" and "FAST" doesn't make sense in this situation as these increments are not increasing tx speed, like on Metamask

src/account/utils/Types.ts
```diff
@@ -210,21 +211,31 @@ export type NonceOptions = {
...
+ * - `1` represents no offset (0% adjustment).
+ * - `2` represents full offset (100% adjustment).
+ * - Values between `1` and `2` indicate partial adjustments.
+ * @public
+ */
+
+/**
+ * Increment gas values by giving an offset, the given value will be an increment to the current estimated gas values, not an override.
+ * @typedef GasOffset
+ * @property {number} [callGasLimitIncrement] - Value used by inner account execution.
```
joepegler, line 227: Can you better describe this? I don't understand what value I should set here from this description

src/account/utils/Types.ts
```diff
@@ -210,21 +211,31 @@ export type NonceOptions = {
...
+ * @property {number} [verificationGasLimitIncrement] - Actual gas used by the validation of this UserOperation.
+ * @property {number} [preVerificationGasIncrement] - Gas overhead of this UserOperation.
+ * @property {number} [maxFeePerGasIncrement] - Maximum fee per gas (similar to EIP-1559 max_fee_per_gas).
```
joepegler, line 230: Are these accurately described? They seem to describe the underlying field, not how it impacts the gas estimate

src/account/utils/Types.ts
```diff
@@ -210,21 +211,31 @@ export type NonceOptions = {
...
+ * An offset value determines the degree of adjustment or displacement.
+ * - `1` represents no offset (0% adjustment).
+ * - `2` represents full offset (100% adjustment).
+ * - Values between `1` and `2` indicate partial adjustments.
+ * @public
+ */
+
+/**
+ * Increment gas values by giving an offset, the given value will be an increment to the current estimated gas values, not an override.
+ * @typedef GasOffset
```
joepegler, line 226: Is 'increment' still the right phrase? Is it a percentage expressed as an integer? If so we should call it 'factor' instead I think. When I hear 'increment' I think 'addition'. When I hear 'factor' I think 'multiplication' (which is actually what is happening right?).

src/account/BiconomySmartAccountV2.ts
```diff
@@ -978,7 +979,7 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
         return this.getPaymasterAndData(partialUserOp, {
           ...paymasterServiceData,
           feeTokenAddress: feeQuote.tokenAddress,
-          calculateGasLimits: true // Always recommended and especially when using token paymaster
+          calculateGasLimits: paymasterServiceData.calculateGasLimits ?? true // Always recommended and especially when using token paymaster
```
livingrockrises, line 982: so they will manually pass false in order to use gas offsets?
will this apply for every use case tx paths?

src/account/BiconomySmartAccountV2.ts
```diff
@@ -1513,21 +1514,62 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
 
-      userOp = await this.getPaymasterUserOp(
-        userOp,
-        buildUseropDto.paymasterServiceData
-      )
-      return userOp
+      if(buildUseropDto.gasOffset){
+        userOp = await this.estimateUserOpGas(userOp)
+
+        const {verificationGasLimitIncrement, preVerificationGasIncrement, callGasLimitIncrement, maxFeePerGasIncrement, maxPriorityFeePerGasIncrement} = buildUseropDto.gasOffset;
```
livingrockrises, line 1520: we should describe well how to provide these
increment compared to which base value?

src/account/BiconomySmartAccountV2.ts
```diff
@@ -1513,21 +1514,62 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+        userOp.verificationGasLimit = toHex(parseInt((Number(userOp.verificationGasLimit ?? 0) * convertToFactor(verificationGasLimitIncrement)).toString()));
+        userOp.preVerificationGas = toHex(parseInt((Number(userOp.preVerificationGas ?? 0) * convertToFactor(preVerificationGasIncrement)).toString()));
```
livingrockrises, line 1522: what would be the cases / negative paths when these values will land up undefined here?

cause in all these cases it will just make resultant 0

src/account/BiconomySmartAccountV2.ts
```diff
@@ -1513,21 +1514,62 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+        userOp.callGasLimit = toHex(parseInt((Number(userOp.callGasLimit ?? 0) * convertToFactor(callGasLimitIncrement)).toString()));
+        userOp.maxFeePerGas = toHex(parseInt((Number(userOp.maxFeePerGas ?? 0) * convertToFactor(maxFeePerGasIncrement)).toString()));
+        userOp.maxPriorityFeePerGas = toHex(parseInt((Number(userOp.maxPriorityFeePerGas ?? 0) * convertToFactor(maxPriorityFeePerGasIncrement)).toString()));
+        
+        userOp = await this.getPaymasterUserOp(
+          userOp,
+          {...buildUseropDto.paymasterServiceData, calculateGasLimits: false},
+        )
+        return userOp
+      } else {
```
livingrockrises, line 1532: why this else path here?
can we somehow optimise this code block

livingrockrises:
was there a strong demand for this feature?
if yes which dapps

src/account/BiconomySmartAccountV2.ts
VGabriel45: calculateGasLimits are set to false in the SDK depending if gasOffsets are use or not

`userOp = await this.getPaymasterUserOp(userOp, {
          ...buildUseropDto.paymasterServiceData,
          calculateGasLimits: false
        })`

src/account/BiconomySmartAccountV2.ts
VGabriel45: Are you talking about the offset values coming from args or about the userOp estimated values ? gasOffset values are handled in convertToFactor for undefined and userOp values are coming from bundler estimation so they should never be undefined or 0

src/account/BiconomySmartAccountV2.ts
VGabriel45: Fixed

tomarsachin2271:
This is an optional feature in the sdk, i personally faced this issue while running some scripts that gas estimations from paymaster, especially callGasLimit was low and was causing internal transaction failure. I had to fallback to using buildUserOp and sendUserOp flow, so i can change these gasValues before signing and sending UserOp.

So while we can handle most of the use case, we can't handle all edge cases. If some dapps is facing this issue, there should be an easy way for them to handle this use case without manually handling userOp on their end.

VGabriel45:
> This is an optional feature in the sdk, i personally faced this issue while running some scripts that gas estimations from paymaster, especially callGasLimit was low and was causing internal transaction failure. I had to fallback to using buildUserOp and sendUserOp flow, so i can change these gasValues before signing and sending UserOp.
> 
> So while we can handle most of the use case, we can't handle all edge cases. If some dapps is facing this issue, there should be an easy way for them to handle this use case without manually handling userOp on their end.

@arcticfloyd1984 Can you confirm if the actual implementation is correct then ? I might just release this in the end if it does no harm but it can be useful to clients. Before releasing I will be removing maxFeePerGas and maxPriorityFeePerGas tho, as increasing this will do no good.

VGabriel45:
> I am good with the implementation @VGabriel45 Also, we should not recommend this to anyone unless necessary especially for mainnets as someone can send at a lower preVerificationGas (we currently accept at a 50% lower value than what we estimate on our end). This can mess up our pricing and make our Bundlers lose money. Please communicate with everyone involved with client integrations that this should be the last resort after infra has checked why the estimates are not sufficient.

Shouldn't be possible to decrease the values using this PR's approach, you can only increase values by a percentage, cannot override them.

livingrockrises:
> > I am good with the implementation @VGabriel45 Also, we should not recommend this to anyone unless necessary especially for mainnets as someone can send at a lower preVerificationGas (we currently accept at a 50% lower value than what we estimate on our end). This can mess up our pricing and make our Bundlers lose money. Please communicate with everyone involved with client integrations that this should be the last resort after infra has checked why the estimates are not sufficient.
> 
> Shouldn't be possible to decrease the values using this PR's approach, you can only increase values by a percentage, cannot override them.

exactly

livingrockrises (approved):

joepegler (approved):
Yeah looks great, nice job

PR #475, opened 2024-04-23
feat: simulateUserOp execution

VGabriel45:
# Summary

Added a method "simulateUserOp()" to simulate user operation execution.

Usecase: 
-  a dapp could use this to warn a user that his user operation will fail before sending it. 
- can also be used for testing

## Change Type

- [ ] Bug Fix
- [ ] Refactor
- [x] New Feature
- [ ] Breaking Change
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Other

# Checklist

---

## PR-Codex overview
This PR adds new functions `Hex` and `BaseError`, updates `simulateUserOp` method, and introduces tests for user operation simulation in `BiconomySmartAccountV2`.

### Detailed summary
- Added `Hex` and `BaseError` imports
- Updated `simulateUserOp` method in `BiconomySmartAccountV2`
- Added tests for user operation simulation in `BiconomySmartAccountV2`

### DIFF
--- a/src/account/BiconomySmartAccountV2.ts
+++ b/src/account/BiconomySmartAccountV2.ts
@@ -1188,44 +1188,12 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
    * @param client The public client.
    * @returns A Promise that resolves to a boolean indicating whether the operation was successful.
   */
-  async simulateUserOp(userOp: UserOperationStruct | Partial<UserOperationStruct>, targetContract: Hex, targetCalldata: Hex, client: PublicClient): Promise<boolean> {
-    const entryPoint = getContract({address: ENTRYPOINT_ADDRESS as Hex, abi: EntryPointAbi, client});
-    // This will always end in catch, as simulateHandleOp() will always revert.
-    const requiredFields: UserOperationKey[] = [
-      "sender",
-      "nonce",
-      "initCode",
-      "callData",
-      "callGasLimit",
-      "verificationGasLimit",
-      "preVerificationGas",
-      "maxFeePerGas",
-      "maxPriorityFeePerGas",
-      "paymasterAndData"
-    ]
-    this.validateUserOp(userOp, requiredFields);
+  async simulateUserOp(txs: Transaction[]): Promise<{success: boolean, message: string}> {
     try {
-      await entryPoint.simulate.simulateHandleOp([{
-        sender: userOp.sender! as Hex,
-        nonce: BigInt(userOp.nonce!),
-        initCode: userOp.initCode! as Hex,
-        callData: userOp.callData! as Hex ,
-        signature: userOp.signature! as Hex,
-        maxFeePerGas: BigInt(userOp.maxFeePerGas!),
-        maxPriorityFeePerGas: BigInt(userOp.maxPriorityFeePerGas!),
-        paymasterAndData: userOp.paymasterAndData! as Hex,
-        preVerificationGas: BigInt(userOp.preVerificationGas!),
-        verificationGasLimit: BigInt(userOp.verificationGasLimit!),
-        callGasLimit: BigInt(userOp.callGasLimit!),
-      }, targetContract, targetCalldata])
-      return false;
-    } catch (error: any) {
-      const baseError = error as BaseError;
-      const metaMessagesArray = baseError.metaMessages![1].split(",");
-      // Remove white spaces and parentheses from each element
-      const cleanedArray = metaMessagesArray.map(element => element.replace(/\s/g, '').replace(/[()]/g, ''));
-      const isSuccessful = cleanedArray[4] === "true"
-      return isSuccessful;
+      await this.buildUserOp(txs)
+      return {success: true, message: "User operation execution passed"}
+    } catch (error) {
+      return {success: false, message: `User operation execution failed: ${error}`}
     }
   }
 
--- a/src/account/BiconomySmartAccountV2.ts
+++ b/src/account/BiconomySmartAccountV2.ts
@@ -1,3 +1,4 @@
+
 import {
   http,
   type GetContractReturnType,
@@ -17,7 +18,8 @@ import {
   parseAbi,
   parseAbiParameters,
   toBytes,
-  toHex
+  toHex,
+  BaseError
 } from "viem"
 import type { IBundler } from "../bundler/IBundler.js"
 import {
@@ -90,6 +92,8 @@ import {
   isValidRpcUrl,
   packUserOp
 } from "./utils/Utils.js"
+import { EntryPointAbi } from "./abi/EntryPointAbi.js"
+import { ENTRYPOINT_ADDRESS } from "../paymaster/utils/Constants.js"
 
 type UserOperationKey = keyof UserOperationStruct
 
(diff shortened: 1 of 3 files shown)


src/account/BiconomySmartAccountV2.ts
```diff
@@ -1173,6 +1177,58 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+   * 
+   * Checks if a user operation will be successful before sending it to the bundler.
+   * 
+   * @param userOp The user operation to simulate.
+   * @param targetContract The target contract to simulate the operation on.
+   * @param targetCalldata The calldata for the target contract.
+   * @param client The public client.
+   * @returns A Promise that resolves to a boolean indicating whether the operation was successful.
+  */
+  async simulateUserOp(userOp: UserOperationStruct | Partial<UserOperationStruct>, targetContract: Hex, targetCalldata: Hex, client: PublicClient): Promise<boolean> {
```
joepegler, line 1191: Can we use getAAError here somewhere? So users know what went wrong with context?

joepegler (commented):
Very nice. Just had one comment / question. Can this return errors with context using getAAError at all?

src/account/BiconomySmartAccountV2.ts
VGabriel45: Good point, I will have a look at this

tests/account/read.test.ts
```diff
@@ -692,4 +694,55 @@ describe("Account:Read", () => {
...
+    })
+
+    const tx = {
+      to: recipient,
+      data: "0x"
+    }
+    const userOp = await smartAccount.buildUserOp([tx])
+
+    const userOpSuccess = await smartAccount.simulateUserOp(userOp, recipient, tx.data as Hex, publicClient);
+    expect(userOpSuccess).toEqual(false);
```
livingrockrises, line 745: why would this fail? wallet doesn't have enough funds?

livingrockrises:
this method is meant to catch execution errors? userop.calldata is wrong as per business logic
and validation errors as well?

can we write more test cases based off this

tests/account/read.test.ts
VGabriel45: Yes, exactly. I have updated the tests to make it clearer.

livingrockrises:
> > this method is meant to catch execution errors? userop.calldata is wrong as per business logic and validation errors as well?
> > can we write more test cases based off this
> 
> entryPoint.simulate.simulateHandleOp will always throw so this is what I'm catching in the try/catch statement

agree, my question was does return info about failed execution or something else too (like AAxx validation problems)

VGabriel45:
> this method is meant to catch execution errors? userop.calldata is wrong as per business logic and validation errors as well?
> 
> can we write more test cases based off this

As per the smart contract method code simulateHandleOp is supposed to simulate both validation and execution of the userOp. In the SDK, as long as the isSuccessful end result value is not true, it means that the userOperation has failed either on validation or execution, the end goal of this utility method is to simulate if an user op would be successfull or not.

livingrockrises:
> > this method is meant to catch execution errors? userop.calldata is wrong as per business logic and validation errors as well?
> > can we write more test cases based off this
> 
> As per the smart contract method code simulateHandleOp is supposed to simulate both validation and execution of the userOp. In the SDK, as long as the isSuccessful end result value is not true, it means that the userOperation has failed either on validation or execution, the end goal of this utility method is to simulate if an user op would be successfull or not.

ok. maybe you can have tests for both
execution failure: paymaster is used to pay gas, but userop.calldata is transferring eth value/erc20 that is not there in the account. (which is somewhat already covered I think)
validation failure: paymaster is used but paymaster doesn't have any deposit on entrypoint (would fail here with AA31)

my advice is to also check again with Yash @arcticfloyd1984  on this whole method usage how it's utilised on bundler and can be utilised

joepegler (commented):
Does this show information from aa-errors?

joepegler:
Looks useful. Could we use this in a 'dev' mode somehow?

VGabriel45:
> Does this show information from aa-errors?

The simulateHandleOp method is supposed to always fail, in the error message there will be a result true/false if the userOp would be executed or not. Not sure if we can use the aa-errors package for this, lmk if you have any idea.

src/account/BiconomySmartAccountV2.ts
```diff
@@ -1173,6 +1177,58 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
     return bundlerResponse
   }
 
+  /**
+   * Simulates a user operation execution.
+   * 
+   * Checks if a user operation will be successful before sending it to the bundler.
+   * 
+   * @param userOp The user operation to simulate.
```
joepegler, line 1185: example please

arcticfloyd1984 (commented):
How is this different from `eth_estimateUserOperation` or `pm_sponsorUserOperation` with `calculateGasLimits: true`?

VGabriel45:
> eth_estimateUserOperation

This method is supposed to simulate the user operation validation + execution, without sending the user operation after the simulation is done, also, it is calling the entrypoint contract directly without a call to the API. We had someone asking for a way to simulate user ops through the SDK without sending them right after (what is happening in our sendUserOp and sendTransaction when passing simulationType flag as "validation" or "validation_and_execution")

arcticfloyd1984:
> > eth_estimateUserOperation
> 
> This method is supposed to simulate the user operation validation + execution, without sending the user operation after the simulation is done, also, it is calling the entrypoint contract directly without a call to the API. We had someone asking for a way to simulate user ops through the SDK without sending them right after (what is happening in our sendUserOp and sendTransaction when passing simulationType flag as "validation" or "validation_and_execution")

Two things:
- You can still expose a method to call `eth_esitmateUserOperationGas` without sending it on chain, but I get it will be slower.
- You can use the [gas estimation package](https://github.com/bcnmy/entry-point-gas-estimations/tree/develop) methods internally to do this. It is a much better way to estimate, especially the execution step. `simulateHandleOp` does not tell you if your `callData` execution passed or not. The gas estimation SDK is written in a way to capture that using state overrides. The only limitation is it will fully work with the networks supporting state overrides in their `eth_call`, for the rest of the networks I have exposed the exact same `simulateHandleOp` method already which gets used in the main `estimateUserOperationGas` by passing a flag. It is already used in production by the bundler and paymaster for months now so it is very well tested.

src/account/BiconomySmartAccountV2.ts
```diff
@@ -1173,6 +1177,58 @@ export class BiconomySmartAccountV2 extends BaseSmartContractAccount {
...
+        callGasLimit: BigInt(userOp.callGasLimit!),
+      }, targetContract, targetCalldata])
+      return false;
+    } catch (error: any) {
+      const baseError = error as BaseError;
+      const metaMessagesArray = baseError.metaMessages![1].split(",");
+      // Remove white spaces and parentheses from each element
+      const cleanedArray = metaMessagesArray.map(element => element.replace(/\s/g, '').replace(/[()]/g, ''));
+      const isSuccessful = cleanedArray[4] === "true"
+      return isSuccessful;
```
joepegler, line 1228: Instead of returning true false let's return:
 
`type SimulationStatus = { status: "true" } | { status: "false"; error: Error }`

Where the error is:

`await getAAError(baseError.message)`

joepegler (changes requested):

tests/account/read.test.ts
TechnoGeek01: I think there should be a test with batched tx's as well.

TechnoGeek01: There should also be a test case with low paymaster balance. It should fail.

PR #576, opened 2024-09-17
Nexus fix: OwnableValidator & OwnableExecutor integration + tests, + add missing features after viem refactor + fix issues & conflicts +++ 

VGabriel45:
## PR-Codex overview
This PR focuses on refactoring the handling of `signer` instead of `holder` in various modules and updates the related tests. It also introduces new functionalities and makes structural adjustments to improve the codebase.

### Detailed summary
- Replaced `holder` with `signer` in multiple files.
- Updated `toHolder` utility to `toSigner`.
- Modified function signatures and types to use `Signer`.
- Added new module `getPreviousModule` for retrieving previous module addresses.
- Enhanced tests for `OwnableExecutor` and `OwnableValidator`.
- Adjusted module installation and uninstalling logic.
- Updated the `package.json` dependencies.
- Changed addresses in `addresses.ts`.

> The following files were skipped due to too many changes: `src/sdk/modules/validators/ownableValidator.test.ts`, `src/sdk/modules/validators/OwnableValidator.ts`


livingrockrises:
conflicts need to be resolved.

src/__contracts/addresses.ts
```diff
@@ -2,9 +2,9 @@
 
 import type { Hex } from "viem"
 export const addresses: Record<string, Hex> = {
-  Nexus: "0x776d63154D2aa9256D72C420416c930F3B735464",
-  K1Validator: "0xd98238BBAeA4f91683d250003799EAd31d7F5c55",
-  K1ValidatorFactory: "0x8025afaD10209b8bEF3A3C94684AaE4D309c9996",
+  Nexus: "0x2ecd86799137FA35De834Da03D876bcc363ec0c3",
```
livingrockrises: when were these deployed and from which branch of nexus? on base sepolia?

src/account/NexusSmartAccount.ts
```diff
@@ -1039,9 +1039,17 @@ export class NexusSmartAccount extends BaseSmartContractAccount {
   async sendUserOp({
     signature,
     ...userOpWithoutSignature
-  }: Partial<UserOperationStruct>): Promise<UserOpResponse> {
-    const userOperation = await this.signUserOp(userOpWithoutSignature)
-    return await this.sendSignedUserOp(userOperation)
+  }: Partial<UserOperationStruct>, userOpSignature?: Hex): Promise<UserOpResponse> {
+    if (!userOpSignature) {
```
livingrockrises: how is this change useful?

src/account/NexusSmartAccount.ts
```diff
@@ -1881,13 +1889,13 @@ export class NexusSmartAccount extends BaseSmartContractAccount {
       | Awaited<ReturnType<typeof this.getInstalledExecutors>>,
     module: Module
   ): Hex {
-    const index = installedModules.indexOf(getAddress(module.moduleAddress))
+    const index = installedModules[0].indexOf(getAddress(module.moduleAddress))
     if (index === 0) {
       return SENTINEL_ADDRESS
     }
     if (index > 0) {
       // @ts-ignore: TODO: Gabi This looks wrong
```
livingrockrises: check this comment. if addressed please delete it

src/account/utils/Types.ts
```diff
@@ -208,6 +208,7 @@ export type BuildUserOpOptions = {
   dummyPndOverride?: BytesLike
   useEmptyDeployCallData?: boolean
   useExecutor?: boolean
+  signatureOverride?: Hex
```
livingrockrises: this would be useful for gas estimation from bundler?
as in when calling eth_estimateUserOpGas

src/modules/utils/Helper.ts
```diff
@@ -152,13 +153,10 @@ export const getUserOpHash = (
   return keccak256(enc)
 }
 
-export const getRandomSigner = (): SignerData => {
+export const getRandomSigner = (): PrivateKeyAccount => {
```
livingrockrises: ?
cc @joepegler

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+    parseAbiParameters
+} from "viem"
+import type { Hex } from "viem"
+import type { NexusSmartAccount } from "../../account/NexusSmartAccount.js"
+import { OWNABLE_VALIDATOR_ADDRESS, SENTINEL_ADDRESS, UserOperationStruct } from "../../account/index.js"
+import { BaseValidationModule } from "../base/BaseValidationModule.js"
+import type { Module } from "../utils/Types.js"
+
+export class OwnableValidator extends BaseValidationModule {
+    public smartAccount: NexusSmartAccount
```
livingrockrises: is it useful for all modules' instances to get SA instance passed?

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+        smartAccount: NexusSmartAccount
+    ) {
+        if (!moduleConfig.data) {
+            throw new Error("Module data is required")
+        }
+        const moduleData = decodeAbiParameters(
+            parseAbiParameters("uint256 threshold, address[] owners"),
+            moduleConfig.data
+        )
+        super(moduleConfig, smartAccount.getSmartAccountOwner())
```
livingrockrises: isn't super supposed to accept a signer?

```
private constructor(moduleConfig: Module, signer: SmartAccountSigner) {
    super(moduleConfig, signer)
  }
```

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+import { BaseValidationModule } from "../base/BaseValidationModule.js"
+import type { Module } from "../utils/Types.js"
+
+export class OwnableValidator extends BaseValidationModule {
+    public smartAccount: NexusSmartAccount
+    public owners: Address[]
+    public threshold: number
+
+    private constructor(
+        moduleConfig: Module,
```
livingrockrises: Module has become a multi-purpose type. have some questions around it, as data can be interpreted multiple ways when it comes to installations and uninstallation if I'm not wrong?

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+        smartAccount: NexusSmartAccount
+    ) {
+        if (!moduleConfig.data) {
+            throw new Error("Module data is required")
+        }
+        const moduleData = decodeAbiParameters(
+            parseAbiParameters("uint256 threshold, address[] owners"),
+            moduleConfig.data
```
livingrockrises: we will need to make some util to make moduleConfig from outside world

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+        this.threshold = Number(moduleData[0])
+        this.owners = [...moduleData[1]] as Address[]
+        this.smartAccount = smartAccount
+    }
+
+    public static async create(
+        smartAccount: NexusSmartAccount,
+        owners: Address[],
+        threshold?: number,
+        hook?: Address
```
livingrockrises: I wonder what is the purpose of this. have noticed in module-sdk as well where a hook is always tied to any kind of a module..

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+        )
+        super(moduleConfig, smartAccount.getSmartAccountOwner())
+        this.threshold = Number(moduleData[0])
+        this.owners = [...moduleData[1]] as Address[]
+        this.smartAccount = smartAccount
+    }
+
+    public static async create(
+        smartAccount: NexusSmartAccount,
+        owners: Address[],
```
livingrockrises: ok I see you already made it. will resolve my above comment

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+        threshold?: number,
+        hook?: Address
+    ): Promise<OwnableValidator> {
+        if (
+            !owners.includes(await smartAccount.getSmartAccountOwner().getAddress())
```
livingrockrises: we're assuming here already initialised smart account. MUST be initialised with K1 as default

livingrockrises: but what if I had passed defaultValidationModule as PasskeyValidator instance in the future? generally how does smartAccount.getSmartAccountOwner() give you answer and fit with .getAddress() later.
just food for thought

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+        ) {
+            throw Error("Signer needs to be one of the owners")
```
livingrockrises: you mean one of the owners should also be an onwer through K1 validator module assumed to be installed.

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+            abi: parseAbi(["function setThreshold(uint256 _threshold) external"]),
+            args: [BigInt(threshold)]
+        })
+
+        const transaction = {
+            to: this.moduleAddress,
+            data: calldata,
+            value: 0n
+        }
+        const userOp = await this.smartAccount.buildUserOp([transaction]);
```
livingrockrises: I see benefit of passing smartAccount instance is userop can be sent right away. in stead of just requesting for calldata

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+    }
+
+    public async isOwner(address: Address): Promise<boolean> {
+        const owners = await this.getOwners();
+        return owners.includes(address);
+    }
+
+    public getDummySignature(): Hex {
+        const dummySignature = "0xe8b94748580ca0b4993c9a1b86b5be851bfc076ff5ce3a1ff65bf16392acfcb800f9b4f1aef1555c7fce5599fffb17e7c635502154a0333ba21f3ae491839af51c";
+        const signatures = Array(this.threshold).fill(dummySignature);
```
livingrockrises: cool

src/modules/validators/OwnableValidator.ts
```diff
@@ -1,179 +1,190 @@
...
+        return concat(signatures) as Hex;
+    }
+
+    async signUserOpHash(userOpHash: string): Promise<Hex> {
+        if (this.isMultiSig()) {
+            throw new Error("Multi-signature required, please pass the multi-signature to the sendUserOp function")
+        } else {
+            return await this.signer.signMessage({ raw: userOpHash as Hex }) as Hex;
```
livingrockrises: what is this.signer here?

tests/modules.ownableValidator.write.test.ts
```diff
@@ -0,0 +1,275 @@
...
+  killNetwork,
+  toTestClient,
+} from "./src/testUtils"
+import type { MasterClient, NetworkConfig } from "./src/testUtils"
+import { createK1ValidatorModule, createOwnableValidatorModule, getRandomSigner } from "../src/modules/index"
+import { OWNABLE_VALIDATOR_ADDRESS } from "../src/account/utils/Constants"
+import { OwnableValidator } from "../src/modules/validators/OwnableValidator"
+import { K1ValidatorModule } from "../src/modules/validators/K1ValidatorModule"
+
+const NETWORK_TYPE: TestFileNetworkType = "PUBLIC_TESTNET"
```
livingrockrises: why resort to public testnet? any particular reason?

livingrockrises (commented):
review i

livingrockrises:
conflicts to be resolved

src/account/utils/Types.ts
VGabriel45: Yes but will remove this for now to simplify the PR, is not needed now, it was making things easier when testing modules with multi signature needed.

src/account/NexusSmartAccount.ts
VGabriel45: It was supposed to skip the "signUserOp" step which gets signature from module instance, a nice usecase for this was for multi signature user ops where the app needs to first get all signatures and encode them into 1 final signature which will then pass here to override the default behaviour of taking sig from module. Will remove this for now.

src/account/NexusSmartAccount.ts
VGabriel45: Let's chat about this on a call

VGabriel45: This is actually how rhinestone is handling it in their sdk also @livingrockrises

src/modules/utils/Helper.ts
VGabriel45: Changed the type to return the whole account object, gives us more control and more utility

src/modules/validators/OwnableValidator.ts
VGabriel45: Yes, since we are building user operations for module transactions

src/modules/validators/OwnableValidator.ts
VGabriel45: We already have access to the signer from the smart account, so nice catch, this is not used and can be removed

VGabriel45: That's what smartAccount.getSmartAccountOwner() is passing

src/modules/validators/OwnableValidator.ts
VGabriel45: You propose to create specific types for each module ? Like instead of having "Module" we would have "OwnableValidatorModuleConfig" ?

src/modules/validators/OwnableValidator.ts
VGabriel45: This is for Kernel modules, Kernel has a hook that is associated with every validator, executor and fallback

src/modules/validators/OwnableValidator.ts
VGabriel45: I don't get this, need more context

src/modules/validators/OwnableValidator.ts
VGabriel45: No, my initial thought was that we need the smart account owner eoa to be part of this owners array too but it does not, I did some changes to fix this.

src/modules/validators/OwnableValidator.ts
VGabriel45: Exactly, we add a new layer of utility to the module instance.

src/modules/validators/OwnableValidator.ts
VGabriel45: this has been updated

tests/modules.ownableValidator.write.test.ts
VGabriel45: fixed

src/account/NexusSmartAccount.ts
livingrockrises: they can call sendSignedUserOp?

src/account/utils/Types.ts
livingrockrises: ok lets take a note.

src/modules/utils/Helper.ts
livingrockrises: is this a custom type as well?

src/modules/validators/OwnableValidator.ts
livingrockrises: ok, my only concern was regarding applying defaultValidatorModule when first time SA instance is created and module will expect SA instance passed.

src/modules/validators/OwnableValidator.ts
livingrockrises: ok I thought by name it returns an address and not signer object.

src/modules/validators/OwnableValidator.ts
livingrockrises: I think it is fine for now if only common properties are mandatory

src/modules/validators/OwnableValidator.ts
livingrockrises: I see

src/modules/validators/OwnableValidator.ts
livingrockrises: specific to nexus account. if default validator module is different then this would not have expected result

src/modules/validators/OwnableValidator.ts
livingrockrises: yup. could be totally unrelated owners

livingrockrises:
ok new conflicts here

src/sdk/account/toNexusAccount.test.ts
```diff
@@ -77,11 +78,10 @@ describe("nexus.account", async () => {
       transport: http()
     })
 
-    nexusClient = await createNexusClient({
-      holder: account,
+    nexusClient = await getTestSmartAccountClient({
```
joepegler: It's a small detail I know, but I really don't like this - just because users need to know what createNexusClient does as it's our most important export. Users need to see it being used across all our tests if possible. There's no need for it to be needlessly obfuscated here. Can we use `createNexusClient` with `fundAndDeploy` instead? afaict they do exactly the same thing.

src/sdk/account/toNexusAccount.ts
```diff
@@ -113,6 +113,9 @@ export type NexusSmartAccountImplementation = SmartAccountImplementation<
     encodeExecute: (call: Call) => Promise<Hex>
     encodeExecuteBatch: (calls: readonly Call[]) => Promise<Hex>
     getUserOpHash: (userOp: Partial<UserOperationStruct>) => Promise<Hex>
+    setActiveValidationModule: (validationModule: BaseValidationModule) => void
+    getActiveValidationModule: () => BaseValidationModule
```
joepegler: Do users need access to the activeValidationModule? I don't think they do, but if they do we don't need a getter function. You can expose activeModule by extending the account:

`activeModule: defaultedActiveModule` [here](https://github.com/bcnmy/biconomy-client-sdk/blob/feat/nexus/src/sdk/account/toNexusAccount.ts#L506).

But do users need this anyway?

src/sdk/account/toNexusAccount.ts
```diff
@@ -113,6 +113,9 @@ export type NexusSmartAccountImplementation = SmartAccountImplementation<
...
+    getSigner: () => Promise<Signer>
```
joepegler: Let's not expose this unless there's a good reason? It's kind of a duplicate because you can access the signer stuff the via: `nexusAccount.client` (it's the masterClient).

src/sdk/account/toNexusAccount.ts
```diff
@@ -202,6 +205,14 @@ export const toNexusAccount = async (
     return _accountAddress
   }
 
+  /**
+   * @description Gets the signer associated with this account
+   * @returns A Promise that resolves to the signer
+   */
+  const getSigner = async (): Promise<Signer> => {
```
joepegler: remove?

src/sdk/account/toNexusAccount.ts
```diff
@@ -345,6 +356,24 @@ export const toNexusAccount = async (
...
+    defaultedActiveModule = validationModule;
+  }
+
+  /**
+   * @description Gets the active validation module for the account
+   * @returns The active validation module
+   */
+  const getActiveValidationModule = (): BaseValidationModule => {
+    return defaultedActiveModule;
+  }
```
joepegler: remove?

src/sdk/account/toNexusAccount.ts
joepegler: Wanna try and keep this file as clean & lean as possible as we'll potentially copy / pasting into viem.

src/sdk/account/toNexusAccount.ts
```diff
@@ -494,12 +523,14 @@ export const toNexusAccount = async (
         parameters
       const address = await getCounterFactualAddress()
       const userOperation = { ...userOpWithoutSender, sender: address }
+      console.log(userOperation, "userOperation");
```
joepegler: log

src/sdk/account/toNexusAccount.ts
```diff
@@ -494,12 +523,14 @@ export const toNexusAccount = async (
...
       const hash = getUserOperationHash({
         chainId,
         entryPointAddress: entryPoint07Address,
         entryPointVersion: "0.7",
         userOperation
       })
+      console.log('calling signUserOpHash');
```
joepegler: log

src/sdk/clients/createNexusClient.test.ts
```diff
@@ -157,19 +157,19 @@ describe("nexus.client", async () => {
     expect(result).toBeTruthy()
   }, 30000)
 
-  test.skip("should create a nexusAccount from an ethers signer", async () => {
-    const ethersProvider = new JsonRpcProvider(chain.rpcUrls.default.http[0])
-    const ethersSigner = new Wallet(pKey, ethersProvider)
+  // test("should create a nexusAccount from an ethers signer", async () => {
```
joepegler, line 156: Why comment instead of skip? Commenting means we have to repeatedly fix imports.

src/sdk/clients/decorators/erc7579/erc7579.decorators.test.ts
```diff
@@ -31,13 +39,13 @@ describe("erc7579.decorators", async () => {
 
     chain = network.chain
     bundlerUrl = network.bundlerUrl
-    account = getTestAccount(0)
+    eoaAccount = getTestAccount(0)
```
joepegler, line 42: Can you change this everywhere if you wanna change it? account seemed fine for me but no biggie

src/sdk/clients/decorators/erc7579/getPreviousModule.ts
```diff
@@ -0,0 +1,84 @@
...
+): Promise<Hex> {
+    const { account: account_ = client.account, module } = parameters
+
+    if (!account_) {
+        throw new AccountNotFoundError({
+            docsPath: "/docs/actions/wallet/sendTransaction"
+        })
+    }
+
+    console.log(parameters, "parameters");
```
joepegler, line 54: log

src/sdk/clients/decorators/erc7579/getPreviousModule.ts
```diff
@@ -0,0 +1,84 @@
...
+        if (!parameters.installedExecutors) throw Error("installedExecutors parameter is missing")
+        installedModules = [...parameters.installedExecutors]
+    } else {
+        throw new Error(`Unknown module type ${module.type}`)
+    }
+
+    return _getModuleByIndex(installedModules, module.address)
+}
+
+function _getModuleByIndex(
```
joepegler, line 70: Is this used anywhere else? Any reason it's in a different function?

src/sdk/clients/decorators/erc7579/uninstallModule.ts
```diff
@@ -67,6 +67,7 @@ export async function uninstallModule<
   }
 
   const account = parseAccount(account_) as SmartAccount
+  console.log(account.address, "account.address")
```
joepegler: log

src/sdk/modules/executors/OwnableExecutor.ts
```diff
@@ -1,198 +1,198 @@
...
-import { type Holder, toHolder } from "../../account/utils/toHolder"
-import type { NexusClient } from "../../clients/createNexusClient"
-import type { Module } from "../../clients/decorators/erc7579"
-import { BaseExecutionModule } from "../base/BaseExecutionModule"
-import type { Execution } from "../utils/Types"
-export class OwnableExecutorModule extends BaseExecutionModule {
-  public nexusClient: NexusClient
-  public owners: Address[]
-  public override address: Address
+// import {
```
joepegler: Can we just delete files that have comments in? You can leave the code in a different branch if you think it'd be useful later

src/sdk/modules/index.ts
```diff
@@ -1,14 +1,14 @@
...
+import { OwnableValidator } from "./validators/OwnableValidator.js"
 import { ValidationModule } from "./validators/ValidationModule.js"
 
 export * from "./utils/Types.js"
 export * from "./utils/Constants.js"
 export * from "./utils/Helper.js"
 export * from "./interfaces/IValidationModule.js"
 
-export const createOwnableExecutorModule = OwnableExecutorModule.create
+// export const createOwnableExecutorModule = OwnableExecutorModule.create
```
joepegler, line 10: Let's stop doing these aliases here. Bit confusing doing it in a different file to where they actually lives. Shall we move each alias in to the relevant file and export them as the default?

src/sdk/modules/validators/k1Validator.test.ts
```diff
@@ -70,57 +70,84 @@ describe("modules.k1Validator.write", async () => {
...
     })
+    console.log(isInstalledBefore, "isInstalledBefore");
 
     if (!isInstalledBefore) {
       const hash = await nexusClient.installModule({
         module: {
           address: addresses.K1Validator,
           type: "validator",
-          context: "0x"
+          data: encodePacked(["address"], [account.address])
```
joepegler, line 87: I think we could expose a `getData` helper from each decorator, because it's likely that the data will be likely very smart-account specific. I'd like for each of our core decorator logic to work with other smart accounts, and a 'getData' helper to construct the relevant information might help with that, so that the main logic remains account agnostic. In this case `const getData = (accountAddress: Address) => encodePacked(["address"], [account.address]);` should be exported from the installModule file, and imported here in the test.

src/sdk/modules/validators/k1Validator.test.ts
```diff
@@ -70,57 +70,84 @@ describe("modules.k1Validator.write", async () => {
...
+        ],
+        [prevModule, encodePacked(["address"], [account.address])]
+      )
+
+      const hashUninstall = nexusClient.uninstallModule({
         module: {
           address: addresses.K1Validator,
           type: "validator",
-          context: "0x"
+          data: deInitData
```
joepegler, line 117: here the 'getData' helper from the uninstallModule file could find the deInitData?

src/sdk/modules/validators/ownableValidator.test.ts
```diff
@@ -0,0 +1,228 @@
...
+  // Test utils
+  let testClient: MasterClient
+  let account: Account
+  let nexusClient: NexusClient
+  let nexusAccountAddress: Address
+  let recipient: Account
+  let recipientAddress: Address
+
+  beforeAll(async () => {
+    network = await toNetwork("FILE_LOCALHOST")
```
joepegler, line 52: Just FYI this is the default setting now. `await toNetwork()` works the same way

src/sdk/modules/validators/ownableValidator.test.ts
```diff
@@ -0,0 +1,228 @@
...
+      throw new Error("Failed to get add owner execution")
+    }
+
+    const owners = await testClient.readContract({
+      address: TEST_CONTRACTS.OwnableValidator.address,
+      abi: parseAbi(["function getOwners(address account) external view returns (address[] ownersArray)"]),
+      functionName: "getOwners",
+      args: [nexusClient.account.address]
+    })
+    console.log(owners, "owners");
```
joepegler: log

src/sdk/modules/validators/ownableValidator.test.ts
```diff
@@ -0,0 +1,228 @@
...
+    })
+    const userOpHash = await nexusClient.sendTransaction({
+      calls: [
+        {
+          to: TEST_CONTRACTS.OwnableValidator.address,
+          data: calldata
+        }
+      ]
+    })
+    console.log(userOpHash, "userOpHash");
```
joepegler: log

src/test/callDatas.js.map
```diff
@@ -0,0 +1 @@
...":["callDatas.ts"],"names":[],"mappings":";;;AAGa,QAAA,cAAc,GAAmC;IAE5D,gBAAgB,EAAE;QAChB,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,kBAAkB;QACxB,OAAO,EAAE,4CAA4C;KACtD;IACD,eAAe,EAAE;QACf,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,iBAAiB;QACvB,OAAO,EAAE,4CAA4C;KACtD;IAED,YAAY,EAAE;QACZ,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,cAAc;QACpB,OAAO,EAAE,4CAA4C;KACtD;IACD,YAAY,EAAE;QACZ,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,eAAe;QACrB,OAAO,EAAE,4CAA4C;KACtD;IACD,eAAe,EAAE;QACf,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,iBAAiB;QACvB,OAAO,EAAE,4CAA4C;KACtD;IACD,eAAe,EAAE;QACf,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,iBAAiB;QACvB,OAAO,EAAE,4CAA4C;KACtD;IACD,gBAAgB,EAAE;QAChB,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,kBAAkB;QACxB,OAAO,EAAE,4CAA4C;KACtD;IACD,gBAAgB,EAAE;QAChB,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,kBAAkB;QACxB,OAAO,EAAE,4CAA4C;KACtD;IACD,qBAAqB,EAAE;QACrB,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,wBAAwB;QAC9B,OAAO,EAAE,4CAA4C;KACtD;IACD,eAAe,EAAE;QACf,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,iBAAiB;QACvB,OAAO,EAAE,4CAA4C;KACtD;IACD,aAAa,EAAE;QACb,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,eAAe;QACrB,OAAO,EAAE,4CAA4C;KACtD;IACD,oBAAoB,EAAE;QACpB,OAAO,EAAE,KAAK;QACd,IAAI,EAAE,sBAAsB;QAC5B,OAAO,EAAE,4CAA4C;KACtD;CACF,CAAA;AAEY,QAAA,kCAAkC,GAC7C,oriCAAoriC,CAAA;AAEzqiC,QAAA,0BAA0B,GACrC,wlhCAAwlhC,CAAA"}
```
joepegler: How did this happen? These all need to be removed 🤔

src/test/testUtils.ts
```diff
@@ -76,6 +73,74 @@ export const getTestAccount = (
...
+ *
+ * @remarks
+ * This function creates a smart account, funds it and deploys it if not already deployed,
+ * and returns the instance. It's useful for setting up test environments
+ * that require a deployed smart account.
+ *
+ * @throws Will throw an error if the smart account deployment fails.
+ */
+
+export const getTestSmartAccountClient = async ({
```
joepegler: Kill it with fire :)

src/test/testUtils.ts
```diff
@@ -76,6 +73,74 @@ export const getTestAccount = (
...
+    index
+  })
+
+  const isDeployed = await smartAccount.account.isDeployed()
+
+  const smartAccountBalance = await getBalance(
+    smartAccount.client as MasterClient,
+    smartAccount.account.address
+  )
+  console.log(smartAccountBalance, "smartAccountBalance")
```
joepegler: log

joepegler (commented):
Great to have install functionality back. Thanks for bringing it back 💪

joepegler:
Just a wider point on our modules. I don't like the folder structure any more. It's difficult to know where to look or why the folders are arranged in the way that they are. I also don't like the Base class stuff that we have now. It feels dated, and I'm not sure it even makes sense given how much variety we will have from our modules. I prefer rhinestone flatter / simplified way of organising modules. Can we replicate something similar to what they have done, paired with decorators that can be extended by a nexusClient instead? I realise this might require a different PR

joepegler:
I've not given this more that a moments thought, but something like...

image

`to`

```
modules >
  smartSessions >
    abi
    constants
    installation
    index
    usage
    decorators
  ownables > 
    abi
    constants
    installation
    index
    usage
    decorators

```

livingrockrises:
checks are failing.

livingrockrises:
> I've not given this more that a moments thought, but something like...
> 
> image
> `to`
> 
> ```
> modules >
>   smartSessions >
>     abi
>     constants
>     installation
>     index
>     usage
>     decorators
>   ownables > 
>     abi
>     constants
>     installation
>     index
>     usage
>     decorators
> ```

folder for each module is ok. most of them will be validators though for which we precisely need these classes because of default and active validator concept.

src/sdk/clients/decorators/erc7579/index.ts
```diff
@@ -51,9 +51,11 @@ import {
 
 export type Erc7579Actions<TSmartAccount extends SmartAccount | undefined> = {
   accountId: (args?: GetSmartAccountParameter<TSmartAccount>) => Promise<string>
-  installModule: (args: InstallModuleParameters<TSmartAccount>) => Promise<Hash>
+  installModule: (args: InstallModuleParameters<TSmartAccount> & { signatureOverride?: Hex }) => Promise<Hash>
   installModules: (
-    args: InstallModulesParameters<TSmartAccount>
+    args: InstallModulesParameters<TSmartAccount> & {
+      signatureOverride?: Hex
```
joepegler, line 61: Can you add signatureOverride to the InstallModulesParameters type instead?

src/sdk/clients/decorators/erc7579/uninstallModule.ts
```diff
@@ -50,14 +50,17 @@ export async function uninstallModule<
   TSmartAccount extends SmartAccount | undefined
 >(
   client: Client<Transport, Chain | undefined, TSmartAccount>,
-  parameters: UninstallModuleParameters<TSmartAccount>
+  parameters: UninstallModuleParameters<TSmartAccount> & {
+    signatureOverride?: Hex
```
joepegler, line 54: add signatureOverride to UninstallModuleParameters instead 👍

src/sdk/clients/decorators/erc7579/index.ts
VGabriel45: Ok I thought is not really related to the module installation, it's related to the user operation so I left it out of that type but sure I can add it if you think is better @joepegler

src/sdk/account/toNexusAccount.test.ts
VGabriel45: I had this previously, before refactor it made sense but now I agree, will modify this

src/sdk/account/toNexusAccount.ts
VGabriel45: Yes there should be a method to fetch the active validation module on a smart account through SDK. 
It's an easy to check what validation module you are going to use before running the next user operation.
Sure I can remove the getter and add the activeModule as an extension, if viem has any issue with this we'll remove it.

livingrockrises: btw I noticed (if we still have similar method like sendTransaction or sendUserop) in module sdk that it passes activeModule instance as a param. so you would not need to "set". both can have its own benefits

VGabriel45:
> getActiveValidationModule

We don't need this, instead we will be migrating to use rhinestone's sdk for module.

src/sdk/account/toNexusAccount.ts
```diff
@@ -178,16 +180,16 @@ export const toNexusAccount = async (
     args: [signerAddress, index, [], 0]
   })
 
-  const defaultedActiveModule =
-    activeModule ??
+  let defaultedActiveModule =
+    activeValidationModule ??
```
livingrockrises, line 184: we can also call this activeValidator
but can keep this as note for now and still merge PR.

livingrockrises (approved):
leaving approved

src/sdk/account/toNexusAccount.ts
VGabriel45: Unless this brakes something I would keep it, otherwise we cannot offer a way to get current active module on the SA through sdk

src/sdk/account/toNexusAccount.ts
VGabriel45: We can add in another PR a param "validationModule" on sendUserOperation and sendTransaction, when this param is passed we use it instead of default active module

PR #585, opened 2024-10-09
fix: extract chain id using URL path instead of regex

wiredmatt:
Hi on behalf of the [Unhosted](https://unhosted.com/) team! We're building our wallet and services on top of this SDK. 
We have contacted your support on Telegram and were advised not to expose the bundlerUrl and paymasterUrl in our client-side application (a chrome-extension wallet like metamask), so our solution is to proxy bundler and paymaster calls through our own backend.

This PR will allow us to keep your SDK compatible with our backend service (and any other team that has this same use case). We can't use the same URL pattern that your regex is checking, as we're handling versioning in a different way (through headers).

In short, we want to be able to instantiate the smartAccount like this:

```
const biconomySmartAccountConfig: BiconomySmartAccountV2Config = {
        signer: signer,
        bundlerUrl: `https://our-backend.com/aa/bundler/${chainId}`,
        paymasterUrl: `https://our-backend.com/aa/paymaster/${chainId}`,
      };
``` 

Above will not work unless the changes proposed in this PR are accepted, as our bundler and paymaster wrapper url don't follow the specified regex.

Thank you.

---

## PR-Codex overview
This PR refactors the `extractChainIdFromUrl` function in `src/bundler/utils/Utils.ts` to simplify its implementation and improve error handling. It also updates related functions to use the new extraction logic.

### Detailed summary
- Replaced regex-based extraction with URL pathname splitting in `extractChainIdFromUrl`.
- Improved error handling for invalid chain IDs.
- Updated `extractChainIdFromBundlerUrl` and `extractChainIdFromPaymasterUrl` to call `extractChainIdFromUrl`.

### DIFF
--- a/src/bundler/utils/Utils.ts
+++ b/src/bundler/utils/Utils.ts
@@ -1,23 +1,21 @@
-export const extractChainIdFromBundlerUrl = (url: string): number => {
+export const extractChainIdFromUrl = (url: string) => {
   try {
-    const regex = /\/api\/v2\/(\d+)\/[a-zA-Z0-9.-]+$/
-    // biome-ignore lint/style/noNonNullAssertion: <explanation>
-    const match = regex.exec(url)!
-    return Number.parseInt(match[1])
-  } catch (error) {
-    throw new Error("Invalid chain id")
-  }
-}
+    const pathSegments = new URL(url).pathname.split("/");
+    const chainId = pathSegments[3];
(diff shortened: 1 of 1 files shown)


VGabriel45:
Hey @wiredmatt, thank you for opening this PR.
I understand your use-case and the reason for doing this but I propose to stick with regex since it is more performant. 
I will open another PR where I'll add this logic with regex and will close this one.

